PR #222, opened 2022-03-21
Change docker base to so_smurf_base

jlashner:
This changes the sodetlib docker from being based on the pysmurf-client docker with SO software built on top, to being built on the SOCS docker with the Smurf/EPICS software manually built in. 

One major reason for this change is that the socs docker is maintained much more regularly than the pysmurf docker, so this will use a more up-to-date version of python and python-packages.

Another reason is that it takes a very long time for spt3g/so3g to build, so it's actually much easier to use a docker with those packages installed and manually install epics.

One thing I'm worried about is the now circular-dependancy between this docker and socs dockers such as the pysmurf-controller docker. I do want socs to be built in here so we can easily use the suprsync-db interface, but there might be a better way to do this. @BrianJKoopman do you have any suggestions? Another possibilty is to copy the general smurf-software build instructions into the pysmurf-controller Dockerfile directly instead of using the sodetlib docker as a base.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ tqdm
 setuptools >= 56.2.0
 ipykernel>=5.5.4
 jedi==0.17.1
-PyYAML==5.3.1
+PyYAML
 numpy
 scipy
 matplotlib


jlashner:
Maybe this should have been on a separate branch since this is a second attempt at this, but in recent commits I switch the sodetlib docker from building spt3g and so3g locally to using the so_smurf_base docker, which is the Smurf docker stack but built on ubuntu 20.04 with python3.8. This has been tested on SLAC's cryostat with the newest version of pysmurf.

BrianJKoopman (approved):
Looks good to me. Why pin the PyYAML version to 5.3.1?

jlashner:
I can't remember exactly why I had to pin it, I remember there was a dependency of some sort where this was the solution. I unpinned it and rebuilt and things seem to be building properly, so it's probably fine to leave with so_smurf_base.

BrianJKoopman:
Gotcha, yeah, sounds good. I tried briefly looking for change logs for v6.0, but didn't find them.

PR #251, opened 2022-06-10
Res fitting

RemingtonGerras:
Re-opening #200 - addressed comments regarding:

1. Changing data product in tune_fit 
2. Added quantile argument for plot_fit_summary to help get rid of outliers.
3. Fixed In-line changes requested 

One thing I do want to point out, I was a little confused with the difference between SMuRF (band, channel) vs (band, resonator_index) - look at `plot_channel_fit` and `get_resfit_plot_txt` . I made sure to match outputs in test cases between the original code and my updated version -  so I believe these functions work as expected.


jlashner (approved):
Hey remy! Thanks for this. Sorry this has taken so long, but finally started trying to use this to create some plots and made some small changes:
 - Changed `channels` field to be the "smurf channel" within a band instead of the absolute channel for consistency across other sodetlib data products
 - Changed `res_index` to be a 1d array with just the res index instead of (band, res_index) since band is already being stored
 - added `res_freqs` field to store center resonance freqs found by tune (generally useful and used in plot functions)
 - Change `S21_mag` to `S21`, and have it store the measured complex transmission for each channel instead of the fitted magnitude of the transmission
 - Added `scan_freqs` so we have the freqs used in the eta scan
 - Changed `plot_channel_fit` so it doesn't take a tune file (all data needed is now stored in the fit_dict)
 - change `plot_channel_fit` so it takes an index instead of (band, channel), which makes it easier to plot individual resonators
 - Change plot text to properly handle unassigned channels
 - Remove x-axis label from histograms, since we don't need those to be in the legend of summary plots

With these changes I'm happy with this module and think we can merge it, though we should submit an issue about adding docs. If you have any problems with changes I made let me know, but if not feel free to merge away!

PR #253, opened 2022-06-10
Move old sodetlib code to legacy dir

jlashner:
Moves the `analysis` and `smurf_funcs` directories to a `legacy` dir since they are not technically supported anymore but still need to be used until we have translator functions for IV files.

This also updates any scripts and modules that import them so they have the correct path.


dpdutcher (approved):
I think I've also moved away from using these versions of the scripts, so I don't mind if they're moved.

PR #254, opened 2022-06-11
c_norm & sc_off tracking, optional base_dir to allow load if differen…

rsonka:
Added saving (during analyze_iv()) of c_norm (phase offset constant) & sc_off (superconducting offset) to IVAnalysis (and documented them). Set _load_am() to automatically initialize those arrays in the class if they didn't. (I'm assuming I don't need to do anything special for axis manager with those?). 

Added an optional argument base_dir to _load_am() to allow loading of IVAnalysis save files if sodetlib is run on a different timestream directory structure, so that I could test my code on Princeton's dedicated analysis computer.

I also did a few minor aesthetic improvements by lining up = signs with spaces.

### DIFF
--- a/sodetlib/operations/iv.py
+++ b/sodetlib/operations/iv.py
@@ -57,14 +57,15 @@ class IVAnalysis:
         at each step (Amps)
     resp : np.ndarray
         Array of shape (nchans, nbiases) containing the squid response 
-        at each bias point, with the sc_off and contextless c_norm corrections
+        at each bias point, multiplied  with the sc_off and contextless 
+        c_norm corrections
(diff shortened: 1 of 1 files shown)


msilvafe:
Hi @rsonka, @jlashner and I are trying to do a new year's cleaning up of open PRs in sodetlib. Would you be able to give us an idea of when you could make the requested changes so we can merge this PR? Let us know if we can help. 

For PEP8 formatting you can use a linter (such as pylint) in whatever IDE you use (I use one in vim and vscode), it'll check the pep8 format and tell you how to fix it. While reading the PEP8 manual is not a bad idea a linter is probably an easier path to getting the formatting changes here fixed. Thanks again for these contributions!

PR #255, opened 2022-06-11
IV Translator functions

jlashner:
Adds generic translators to translate between IVAnalysis objects and the old "pysmurf" dict formats where the dict is indext by:
```
d[band][channel]
```
Going from old format --> new format will inevitably leave out some data that isn't present in the old format, but it is enough for simple analyses, and I have tested it with the plotting functions in the iv module and it seems to work fine.

This also moves much of the array initialization from __init__ to a separate function that can be called if you know the number of channels and number of biases you are stepping through.

### DIFF
--- a/sodetlib/operations/iv.py
+++ b/sodetlib/operations/iv.py
@@ -119,26 +119,32 @@ class IVAnalysis:
             self.bias_groups = self.run_kwargs['bias_groups']
 
             am = self._load_am()
-            self.nchans= len(am.signal)
-
+            self.nchans = len(am.signal)
             self.bands = am.ch_info.band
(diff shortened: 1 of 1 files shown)


rsonka:
Hate that I apparently missed this. This doesn't currently change the units to original sodetlib, which stored current in μA, volts in μV, and powers in pW. (if this is trying to convert to a pysmurf file it should also have a field called 'trans idxs' that is IVAnalysis's idxs without the middle p_sat one. 

I also have a problem where I've only been given the iv_analysis file, not the raw data, from the field (the person I'm talking to can't find the raw data.), and this doesn't help in that situation, as I understand it.

Issue #217, opened 2022-03-11
session_id occasionally returned as 0


dpdutcher:
Traceback:
```
2022-03-11T18:06:43+0000 ufm_biasstep_sodetlib.py: Traceback (most recent call last):
  File "/sodetlib/sodetlib/operations/bias_steps.py", line 876, in take_bias_steps
    bsa.run_analysis(save=True, **_analysis_kwargs)
  File "/sodetlib/sodetlib/operations/bias_steps.py", line 265, in run_analysis
    self._load_am(arc=arc)
  File "/sodetlib/sodetlib/operations/bias_steps.py", line 316, in _load_am
    self.am = sdl.load_session(self.meta['stream_id'], self.sid)
  File "/sodetlib/sodetlib/stream.py", line 94, in load_session
    base_dir=base_dir)
  File "/sodetlib/sodetlib/stream.py", line 36, in get_session_files
    os.path.join(subdir, f) for f in os.listdir(subdir)
FileNotFoundError: [Errno 2] No such file or directory: '/data/so/timestreams/0/crate1slot2'
```

I've run into this problem a few times now, in the context of running bias steps on three modules at once via OCS, running sodetlib scripts in the pysmurf-controller image. I suspect the issue is due to the lines below, and that the number of attempts or the wait time needs to be increased. I doubled the attempts to 10 and it still had a failure, and I'm now attempting increasing the wait time at each step to 0.5s. I think this is symptomatic of a bigger issue, though.

https://github.com/simonsobs/sodetlib/blob/60592c8d34b32a7d4c0899438df141465644c0a9/sodetlib/stream.py#L162-L169

dpdutcher:
Of course, the day I close this, it happens again.

jlashner:
Unfortunately this issue is a bit difficult to debug without a way to reliably reproduce. I think a good next step to give us more info is to modify the Smurf streamer to track how long exactly it's taking for the session-id to be set, then we can see if that's actually the issue and work on minimizing that time if it is

jlashner:
Can you link to the paths to the generated g3 files on simons1?

dpdutcher:
Ah, it wasn't during IVs actually, but a timestream-taking step in between IVs. More info, for posterity: 
```
#take 30s timestream for noise
sid = sdl.take_g3_data(S, 30)
am = sdl.load_session(cfg.stream_id, sid, base_dir=cfg.sys['g3_dir'])
```

[...]

PR #260, opened 2022-07-22
increase attempts and sleep time for getting session_id

dpdutcher:
Addresses #217 
Because it's been 4 months with no better solution and this came up again in LATR testing.

Since implementing this I don't think I've ever seen session_id returned as 0.


jlashner (approved):
I think it's pretty strange that it's taking so long for session-id to be set in the streamer since status frames aren't buffered at all. If you really think this helped with the problem I'm not opposed to merging this though.

PR #261, opened 2022-07-25
small bugfixes for plotting

jlashner:
A couple of small bugs that I found when trying to run plotting functions using saved data.


msilvafe (approved):
All looks good, thanks for testing offline plotting.

PR #263, opened 2022-07-26
Jackhammer updates

jlashner:
Updates to jackhammer made during Penn LATR setup:

- Changes `dump` fucntions so they don't dump entire rogue tree by default (this is time-consuming and fails when epics servers aren't up). Adds argument that can be set via command line if you actually want rogue-tree dumps
- Adds `deactivate` and `activate` sub-commands to deactivate and activate slots on the smurf-crate

### DIFF
--- a/hammers/jackhammer
+++ b/hammers/jackhammer
@@ -521,16 +521,18 @@ if __name__ == '__main__':
                              help="If true, will attempt to connect to pysmurf smurf and dump the rogue tree")
     dump_parser.set_defaults(func=dump_func)
 
-    ########### Jackhammer dump parser ###########
-    dump_parser = subparsers.add_parser('deactivate', help='Deactivates slots')
-    dump_parser.add_argument('slots', nargs='*', type=int,
+    ########### Jackhammer deactivate parser ###########
+    deactivate_parser = subparsers.add_parser('deactivate', help='Deactivates slots')
+    deactivate_parser.add_argument('slots', nargs='+', type=int,
                              help='Specifies the slots to deactivate')
-    dump_parser.set_defaults(func=deactivate_func)
+    deactivate_parser.set_defaults(func=deactivate_func)
 
-    dump_parser = subparsers.add_parser('activate', help='Activates slots')
-    dump_parser.add_argument('slots', nargs='*', type=int,
+    ########### Jackhammer activate parser ###########
+    activate_parser = subparsers.add_parser('activate', help='Activates slots')
+    activate_parser.add_argument('slots', nargs='+', type=int,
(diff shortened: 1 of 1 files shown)


hammers/jackhammer
```diff
@@ -501,8 +517,20 @@ if __name__ == '__main__':
...
+    ########### Jackhammer dump parser ###########
+    dump_parser = subparsers.add_parser('deactivate', help='Deactivates slots')
+    dump_parser.add_argument('slots', nargs='*', type=int,
+                             help='Specifies the slots to deactivate')
+    dump_parser.set_defaults(func=deactivate_func)
+
+    dump_parser = subparsers.add_parser('activate', help='Activates slots')
+    dump_parser.add_argument('slots', nargs='*', type=int,
+                             help='Specifies the slots to activate')
+    dump_parser.set_defaults(func=activate_func)
```
msilvafe, line 533: A few questions...
1) Why is the activate/deactivate slots part of the "dump_parser" instead of "hammer_parser" or its own section in here?
2) What is the default action if you don't pass it a slot or list of slots? 

I think if my first question doesn't matter and the 2nd one you fix by making it fail in a reasonable way then I'm happy to accept and merge.

msilvafe (changes requested):

hammers/jackhammer
```diff
@@ -501,8 +517,20 @@ if __name__ == '__main__':
...
     dump_parser.set_defaults(func=dump_func)
 
+    ########### Jackhammer dump parser ###########
+    dump_parser = subparsers.add_parser('deactivate', help='Deactivates slots')
+    dump_parser.add_argument('slots', nargs='*', type=int,
+                             help='Specifies the slots to deactivate')
+    dump_parser.set_defaults(func=deactivate_func)
+
+    dump_parser = subparsers.add_parser('activate', help='Activates slots')
+    dump_parser.add_argument('slots', nargs='*', type=int,
```
dpdutcher, line 531: Should these be nargs='+' , so `slots` is a required argument? As-is, if not provided with any arguments this will drop you into the command line on the shelf manager, which might be confusing

hammers/jackhammer
jlashner: 1. Good catch. Turns out the variable name of the parser doesn't actually matter since I'm creating a new one, but I will rename to match current style
2. I'll take Daniel's suggestion and change nargs to '+'

jlashner:
Thanks for the feedback! Just submitted a fix addressing comments, so if you could re-review that would be great

msilvafe (approved):
Thanks for addressing comments.

PR #264, opened 2022-07-26
squid_curve changes for easier offline plotting

jlashner:
Small changes to squid-fit functions that were required for easier offline analysis + plotting. Mainly I made it so that the analysis function will update the squid data dict with fit params, when before that was only happening in the run-function.


RemingtonGerras (approved):
Yup, changes look good. Thanks for handling that!

Issue #266, opened 2022-07-27
take_bgmap fails with all NaN array encountered


dpdutcher:
https://github.com/simonsobs/sodetlib/blob/d62faa64343e4931db5442466ac9aeea23bcecb3/sodetlib/operations/bias_steps.py#L542

Debugging is hard because no object is returned; should catch this and return something more useful.

jlashner:
I thought the bias step analysis object is still return and can be further inspected: https://github.com/simonsobs/sodetlib/blob/d62faa64343e4931db5442466ac9aeea23bcecb3/sodetlib/operations/bias_steps.py#L1142-L1152

Is that not the case?

jlashner:
Ah maybe its not returning because it's trying to plot the bg assignment: https://github.com/simonsobs/sodetlib/blob/d62faa64343e4931db5442466ac9aeea23bcecb3/sodetlib/operations/bias_steps.py#L1014-L1015

We should check if it actually finished the analysis before plotting

PR #267, opened 2022-07-28
only plot bg_assignments if analysis successful

dpdutcher:
Addresses #266 

Not actually tested, but a straightforward change. Just checks `hasattr(bsa, 'bgmap')` before plotting.


jlashner:
Hi Daniel, just added a small commit that additionally addresses what I think is the underlying problem causing the exception in the first place. Tested at Penn and seems to be working now.

jlashner (approved):
Tested and ready to merge!

PR #269, opened 2022-08-10
In iv_analysis, clean input arrays of nans before polyfit

### LINKED ISSUE #268: NaNs ruin polyfits in analyze_iv
https://github.com/simonsobs/sodetlib/blob/c1acd700e161d75f9e75980dbbb9a3115e7b373c/sodetlib/operations/iv.py#L364-L368

If iva.resp[i] contains a `nan`, these polyfits will return `[nan. nan]`, and then all of iva.resp[i] becomes `nan`. 
In my testing, I've seen that if one detector has a NaN, then every detector on that bias line has a NaN at that same index. So this issue drops whole bias lines from iv results.

Relatively simple fix, using np.isfinite to index the input arrays to the polyfits. PR incoming.

dpdutcher:
Resolves #268 

I tested this by re-analyzing an old IV, and I recovered a bias line's worth of data that had been dropped.
I believe this is the main culprit of why I was getting different and varying yields from newer sodetlib IVs compared to older versions.

### DIFF
--- a/sodetlib/operations/iv.py
+++ b/sodetlib/operations/iv.py
@@ -361,11 +361,16 @@ def analyze_iv(iva, psat_level=0.9, save=False, update_cfg=False):
             continue
         iva.idxs[i, 1] = nb_idx
         nb_fit_idx = (iva.nbiases + nb_idx) // 2
-        norm_fit = np.polyfit(iva.i_bias[nb_fit_idx:],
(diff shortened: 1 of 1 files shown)


jlashner (approved):
This seems like a good check to have. Do you happen to know how NaN's are making their way in?

Issue #262, opened 2022-07-26
UXM Setup pulls attens from dev cfg


jlashner:
Currently UXM setup will only run estimate attens if 'uc_att' or 'dc_att' is None in the device config. If those are not None it will not update attenuation values, which is kind of unintuitive, and caused some issues when setting up at Penn.

I think more reasonable behavior is to change the uxm-setup function to not pull any values (except for target drain currents etc.) from the device config, and instead assume everything is unoptimized. Then we can add kwargs to skip steps such as estimating atten / finding gate voltages if we want to run with pre-loaded values.

PR #270, opened 2022-08-11
add ability to choose if we want to reload attens or not. 

kmharrington:
solves #262

### DIFF
--- a/sodetlib/operations/uxm_setup.py
+++ b/sodetlib/operations/uxm_setup.py
@@ -414,9 +414,10 @@ def uxm_setup(S, cfg, bands=None, show_plots=True, update_cfg=True,
     update_cfg : bool
         If true, will update the device cfg and save the file.
     modify_attens : bool
-        If true, will run estimate_uc_dc_atten to find a set of
-        attenuations that will work for estimate_phase_delay and
-        find_freq.
+        If true or attenuations are not set in the device config, will run
+        estimate_uc_dc_atten to find a set of attenuations that will work for
+        estimate_phase_delay and find_freq. If False and attenuation values
+        already exist in the device config, this will use those values.
     skip_phase_delay : bool
         If True, will skip the estimate_phase_delay step
     skip_setup_amps : bool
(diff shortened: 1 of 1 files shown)


jlashner (commented):
Looks good to me, but I think if `modify_attens` is True, and there are atten values saved in the device config, it makes sense to me for uxm_setup to set attens to whatever's in the device cfg instead of just leaving it.

kmharrington:
you mean if modify_attens is False? 

If True --> run attenuation optimization. If False --> load from config file.

kmharrington:
@jlashner  I think this is what you're looking for?

jlashner:
I think how this is implemented it'll change the attens to the dev-config values if `modify_attens` is False. One sec... about to commit somethingg

jlashner:
@kmharrington can you check that commit and see if you agree?

dpdutcher:
I think there are different understandings of "modify_attens":
1. In one sense, apply possibly different attens to whatever the smurf board is currently using, but that could match the device config. (modify what is currently in use).
2. In the other sense, do not just use the settings from the config file, find appropriate ones from scratch again (modify what is in the config file).

kmharrington:
Yeah Daniel, you called the confusion. At LATR we want a kwarg that will force the re-estimation of the attenuation

kmharrington:
So, sounds like we want two kwargs? maybe `force_estimate_attens` and `load_config_attens`?

dpdutcher:
I like the version as of https://github.com/simonsobs/sodetlib/pull/270/commits/598060a6eef1f1430d35fc5f02f1327c9d069fd8
> I think how this is implemented it'll change the attens to the dev-config values if modify_attens is False.
What would the attens be set to if not what's in the dev-config? (okay, technically what's in the pysmurf config file, but no one looks at that). This is uxm_setup after all, first time set up, so it doesn't really matter what's on the board, right?

jlashner:
If you've done work to set attens manually and trust whatever attens are already loaded into the smurf then you may want to run uxm_setup without touching the already loaded values. Probably not from the pysmurf-config, but from general operation and interactive debugging. I've definitely been in that situation

jlashner:
I support second kwarg to force estimate attens, and think `modify_attens=False` should mean don't touch currently loaded values

jlashner:
But also we should update the docstrings to reflect whatever our choice is

jlashner:
Reviewing this thread and now I'm agreeing more with Katie's definition of what `modify_attens` should mean. I reverted back to Katie's commit and added some clarification in the docs.

jlashner (approved):

PR #271, opened 2022-08-11
Add default option for output dir

kmharrington:
I don't actually have a way to test this, but an easier "cfg.dump_configs()" would be nice to have.


jlashner (approved):
Looks good to me!

Issue #274, opened 2022-08-16
iv_analysis returns NaN for whole bias group R_n, p_sat.


dpdutcher:
P_sat calculation is skipped for channels with NaN R_n, and R_n is calculated using np.mean.
Similar to #268 , I've seen NaNs in iva.resp occur across a whole bias line, and the R_n calculation will then make several other quantities NaN.

Going to replace all calls to np.mean to np.nanmean.

jlashner:
Closed w/ #275

PR #275, opened 2022-08-16
Replace all np.mean with np.nanmean

dpdutcher:
To address #274 and other issues bound to arise in the future.


jlashner (approved):
Merge away!

PR #279, opened 2022-09-02
Load pcie fw

jlashner:
Adds some scripts / docs making it easier for people to download and install pcie fw


dpdutcher (approved):
Yuhan said these instructions work so I guess I'll approve!

PR #285, opened 2022-09-06
Issue238

jlashner:
Adds current version / hash info to `get_metadata` output


msilvafe:
Great, this looks like it directly addresses the issue I submitted. Thanks for adding.

PR #295, opened 2022-09-17
Grad descent function

jlashner:
This PR adds two functions, described below. I think people should review this but I won't merge it in until the uxm_setup_changes branch is cleaned up and merged so Penn doesn't have to switch back and forth between branches.

## run_grad_descent_and_eta_scan

This is a wrapper around smurf's gradient descent and eta scan functions, and critically allows grad-descent parameters to be set using the device config.

When I was looking at tuning issues at Penn I found a few interesting things:
- The default `maxIters` parameter used by gradient is 15, which is way too small. It is not able to recover from shifts in the resonance frequency, unless the `gain` parameter is tuned really well for that particular system. Increasing this to ~100 or 150 makes gradient descent _much_ more consistent and much better at recovering from frequency error, even without touching other parameters
- Some very reasonable combinations of parameters break gradient descent in such a way that it turns off all tones and prevents you from running any other scan operations....

`run_grad_descent_and_eta_scan` will take grad-descent parameters from the device cfg, and will try recover from grad-descent errors when they happen. Additionally, if passed the `updated_tune` kwarg, it will update the smurf-tune object and write a new one with the new freqs and eta parameters.

Based on the tests run at penn yesterday I'm hopeful that this will solve tune-stability issues by making relock w/o setup notches work better.

## plot_channel_resonance

This takes a band and channel and will measure the freq response around just that channel, and present a plot like this:
resonance
which informs you of how close smurf's probe tone is to the actual resonance, and whether the eta parameter is correct. This is a powerful debugging tool, since it can point you to why a channel lost tracking or has high noise.

In the image above the curve looks kinda noisy because flux-ramp is on, but if you turn it off it will be identical to the data taken in setup-notches.

### DIFF
--- a/sodetlib/det_config.py
+++ b/sodetlib/det_config.py
@@ -66,7 +66,7 @@ exp_defaults = {
 
     # Tracking stuff
     "flux_ramp_rate_khz": 4, "init_frac_pp": 0.4, "nphi0": 5,
-    "f_ptp_range": [10, 200], "df_ptp_range": [0, 50], "r2_min": 0.9,
+    "f_ptp_range": [10, 200], "df_ptp_range": [0, 100], "r2_min": 0.9,
     "min_good_tracking_frac": 0.8,
     'feedback_start_frac': 0.02, 'feedback_end_frac': 0.98,
 
@@ -81,6 +81,15 @@ band_defaults = {
 
     # Band-specific tracking stuff
     "lms_gain": 0, "feedback_gain": 2048, "frac_pp": None, "lms_freq_hz": None,
+
+    # Gradient Descent Parameters
+    "gradientDescentMaxIters": 150,
+    "gradientDescentAverages": 2, 
+    "gradientDescentGain": 0.001,
+    "gradientDescentConvergeHz": 500,
+    "gradientDescentStepHz": 5000,
+    "gradientDescentMomentum": 1,
+    "gradientDescentBeta": 0.1
 }
 bg_defaults = {}
 
--- a/sodetlib/operations/uxm_relock.py
+++ b/sodetlib/operations/uxm_relock.py
@@ -4,6 +4,7 @@ import sodetlib as sdl
 from sodetlib.operations.tracking import relock_tracking_setup
 from sodetlib.operations import uxm_setup
 
+import matplotlib.pyplot as plt
 
 @sdl.set_action()
 def reload_tune(S, cfg, bands, setup_notches=False,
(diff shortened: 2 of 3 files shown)


msilvafe:
> In the image above the curve looks kinda noisy because flux-ramp is on, but if you turn it off it will be identical to the data taken in setup-notches.

I'm a bit surprised that this works with the flux ramp on, it should smear out your resonance response making it harder to identify the real resonance frequency. Also I thought that when `set_run_serial_find_freq` is called it puts tracking back in the "integral" mode i.e. it only does slow tracking of the DC term not all of the harmonics of the `lms_freq`, did you confirm that tracking was fine after this and that none of those registers (like lmsEnable1,2,3, etc.) got toggled off? Is the plan in the long run to turn off the flux ramp, run this then rerun tracking setup as part of an updated relock function or is there a mode you envision where we run this only without shutting off the flux ramp.

jlashner:
I agree serial_gradient_descent should always be run with flux-ramp off. Where do you see the thing about "integral" mode? I can only find it disabling feedback entirely, but nothing about the lmsEnable 1/2/3 vars. But yes, I think the plan should be to turn off flux-ramp, re-run grad descent and eta scan with this function, re-setup tracking.

For debugging purposes, I think the `plot_channel_resonance` function that basically tells you how well the res-freq and eta param is set should work without needing to disable flux-ramp, which is the context I was in when I made the plot above.

msilvafe:
> I agree serial_gradient_descent should always be run with flux-ramp off. Where do you see the thing about "integral" mode? I can only find it disabling feedback entirely, but nothing about the lmsEnable 1/2/3 vars. But yes, I think the plan should be to turn off flux-ramp, re-run grad descent and eta scan with this function, re-setup tracking.
> 
> For debugging purposes, I think the `plot_channel_resonance` function that basically tells you how well the res-freq and eta param is set should work without needing to disable flux-ramp, which is the context I was in when I made the plot above.

Ok ya I understand the application and agree this is a useful debug function. I didn't explicitly see that lms1,2,3 get disabled but thought maybe the rogue call to turn on the serial find freq step may toggle the other registers too but sounds like that's not the case. I was just recalling the state of the system at different points as you tune up from everything off (and integral tracking isn't a named thing in the code just when it's in the state that between having run setup_notches and tracking_setup).

jlashner:
I'm going to merge this bc we really want it for P10R1. Any bugfixes found during testing will go in the P10R1 branch

msilvafe (commented):
this is trust.

msilvafe (approved):

PR #296, opened 2022-09-28
New tool to check on which LNAs seem to be working

jeffiuliano:


jeffiuliano:
I tested this with random full_band_resp data on simons1 (so I know the fn runs and displays what I want), but I can't test it actual LNAs at the moment.  Maybe SAT can give it a try.  Also wasn't sure 1) if we wanted to only look at some frequency range of the full_band_resp, or if the raw adc count fn is better, 2) if we needed/wanted the set_action decorator for all these functions.

sodetlib/operations/power_amplifiers.py
```diff
@@ -0,0 +1,163 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import sodetlib as sdl
+
+sdl.set_action()
```
jlashner, line 5: This needs to be
```
@sdl.set_action()
```
to act as a decorator. Also this really only needs to be added to functions that produce archived data files.

sodetlib/operations/power_amplifiers.py
```diff
@@ -0,0 +1,163 @@
...
+    Returns
+    ----------
+    metric: flot
+    	Median value of the fft returned by full_band_resp
+    data: array
+		(f, resp) -- Raw data returned by full_band_resp
+	"""
+	f, resp = S.full_band_resp(band=b)
+	# Or do we need median of some frequency range???
+    return np.nanmedian(np.real(resp)), (f,resp)
```
jlashner, line 26: White space issues, this needs to be fixed throughout. For docs formatting / white space I'd reference other sodetlib operations

jlashner:
Oops I didn't mean to submit that as a review, I was actually trying to cancel it lol. This looks really good, though it seems like there's a lot that still needs to be tested. I can commit changes to this as I test it on SAT1 and then have you review my modifications if that works for you!

dpdutcher:
I had suggested making a new amplifiers module, which would eventually also house the amplifier functions currently in uxm_setup, but I also like the idea of a quality control module. The loopback tests currently don't have a good place to live otherwise.

jeffiuliano:
That sounds good to me!  Definitely needs to be tested on actual hardware.

jeffiuliano:
Fixed a few of the typos I missed that Jack found.  Also the whitespace (apparently I had tabs instead of spaces in some lines)

jeffiuliano:
I wasn't sure where we wanted the action decorator, feel free to remove that from those fns when you test if if it's not needed.

PR #298, opened 2022-10-05
Do not fit for tau if response to bias steps is all nans

### LINKED ISSUE #297: bias step analysis fails if not all bias groups have data
While `take_bias_steps` accepts a `bgs` argument, this is not passed to `run_analysis`, and the inner function `_fit_tau_effs` is hardcoded to run over 12 bias lines. `curve_fit` can then fail with the error

ValueError: `ydata` must not be empty!

I recommend passing the `bgs` argument to run_analysis and _fit_tau_effs.

dpdutcher:
Resolves #297 

There were many ways to go about fixing this; this way seemed the least intrusive and most in-line with the existing code.

### DIFF
--- a/sodetlib/operations/bias_steps.py
+++ b/sodetlib/operations/bias_steps.py
@@ -763,6 +763,8 @@ class BiasStepAnalysis:
             if not len(rcs):
                 continue
             ts = self.resp_times[bg]
+            if not len(ts[~np.isnan(ts)]):
+                continue
             for rc in rcs:
                 resp = self.mean_resp[rc]
                 m = (ts > tmin) & (~np.isnan(resp))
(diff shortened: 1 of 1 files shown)


dpdutcher:
Tested on Princeton system with UFM-Mv27, working well.

msilvafe:
For context can you explain when you get nans out of the bias step code?

dpdutcher:
If you don't run bias steps over all bias groups, whichever bias groups you omitted will have empty values for `edge_idxs` and `edge_signs`, and nan entries in `resp_times`, `step_resp`, and `mean_resp`, as well as the DC parameters.

I think most of those are fine, the only issue I was having is when it would try to fit an all-NaN array for tau, and then it would crash with a ValueError. I could either catch that ValueError, or just not run tau fitting on bias groups that have all NaNs in their `resp_time`.

msilvafe (approved):
Ok makes sense and your fix looks good. I'll go ahead and merge!

PR #311, opened 2022-10-26
Add g3 tagging based on data packaging plans

kmharrington:
Starts solving #309 for the most used operations that create data streams. I don't have a system I can currently test this on but would love to see this in action for p10r1. Could someone do a quick test and merge?

### DIFF
--- a/sodetlib/noise.py
+++ b/sodetlib/noise.py
@@ -473,6 +473,7 @@ def take_noise(S, cfg, acq_time=30, plot_band_summary=True, nbins=40,
                show_plot=True, save_plot=False, plotted_rchans=None,
                wl_f_range=(10,30), fit=False,
                nperdecade=10, plot1overfregion=False, save_dir=None,
+               g3_tag=None,
                **asd_args):
     """
     Streams data for specified amount of time and then calculated the ASD and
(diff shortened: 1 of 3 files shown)


sodetlib/noise.py
```diff
@@ -540,7 +543,9 @@ def take_noise(S, cfg, acq_time=30, plot_band_summary=True, nbins=40,
     if save_dir is None:
         save_dir = S.plot_dir
 
-    sid = sdl.take_g3_data(S, acq_time)
+    if g3_tag is None:
+        g3_tag = "oper,noise"
```
jlashner, line 547: Do you imagine the g3_tag var to be set by users to describe what they're doing? Like adding a tag "transition_with_hwp" to describe a noise timestream?

If so maybe we should change this so it appends to the tag list instead of overwriting it. If we decide to do this the tags should probably be stored in as a set until it's time to write them out to avoid duplicating the 'oper' tag.

jlashner (commented):
Question inline about implementation and passing of tags, but the rest looks good. Is there any documentation of how tags are used in the pipeline? If we overwrite g3_tag with a custom tag is it going to break downstream?

kmharrington:
documentation is in the flow to librarian doc here (section 5): https://github.com/simonsobs/tod2maps_docs 
latest compiled here: https://simons1.princeton.edu/docs/tod2maps_docs/flow_to_librarian/flow_to_librarian_LATEST.pdf

I went back and forth about overriding vs appending but I think overriding will work out better in the end. Ex: overriding is already necessary with bgmap vs bias steps. These tags shouldn't actually come from humans when we're on-site. That doc lays out that we expect them to come from sorunlib through the pysmurf-controller in the end. 

I was also less excited about adding an append option _and_ an over-ride option. I think those would basically only be used in testing but obviously no one is adding tags to the operations right now because it hasn't been an option before.

jlashner:
Makes sense to me! Thanks for this.

Issue #313, opened 2022-11-12
jackhammer --soft should not run S.setup()


dpdutcher:
https://github.com/simonsobs/sodetlib/blob/3155cc4cd4e880bf9b3b26c7723dcd757066b469/hammers/jackhammer#L488-L495

Shawn declares that `setup` should only be run once per board power cycle, but the default `--soft` mode of jackhammer hammer runs `setup` without power cycling the boards. I usually pass the `--skip-setup` option if I'm only soft-hammering because the dockers crashed, but I think that option should get applied automatically.
In the LATR testing they were soft hammering quite a bit, without skipping the setup, and the (incorrect) amplifier bias values from their pysmurf configs were getting applied and actually seemd to cause some heating. That's what prompted my pysmurf issue about that here: https://github.com/slaclab/pysmurf/issues/724.

PR #314, opened 2022-11-14
jackhammer only runs setup if boards are rebooted

dpdutcher:
`S.setup()` should only be run once per board power cycle. This prevents the user from doing otherwise. If actually needed, setup can later be run independently of hammering.

Addresses #313

### DIFF
--- a/hammers/jackhammer
+++ b/hammers/jackhammer
@@ -378,7 +378,7 @@ def hammer_func(args):
         epics_server = f'smurf_server_s{slot}'
         check_epics_connection(epics_server, retry=True)
 
-    if not args.skip_setup:
+    if reboot and not args.skip_setup:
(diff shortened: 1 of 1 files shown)


jlashner:
Have you tested this? I wasn't aware until recently that we weren't supposed to be running setup for a soft-reset, so I haven't tested this operation mode / am not 100% all the software variables will be set correctly without running setup. If you've ran soft-reset without setup and everything works normally then I am happy to merge.

dpdutcher:
I have tested this in the following use case:
I got the "RuntimeError: epics failed to respond" error, and I couldn't communicate with the boards. I then did `jackhammer hammer --soft --skip-setup` and jumped right back into the smurf session once the dockers were back up. It was as if nothing had happened. No relocking or anything required.

jlashner (approved):
Works for me!

PR #316, opened 2022-11-23
Update ocs version within Docker image

BrianJKoopman:
For the pysmurf-controller we need the newest version of ocs. I made a Python 3.6 branch for ocs based on the latest develop branch, so that'll have the latest changes (and need to be kept up to date in the future.) We're doing the same thing on socs at the moment. This installs that new py36 ocs branch, and also checks out spt3g before Python detection in 3.6 broke.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ RUN apt-get install -y \
     libopenblas-dev \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


jlashner (approved):
Looks good, though we should fix this in the smurf dockers ASAP, maintaining a separate py36 branch seems like a nightmare

PR #317, opened 2022-11-28
Fix docs builds in Actions workflows

BrianJKoopman:
I noticed these were failing for #316 and thought I'd fix.

pyfftw installation is failing for some reason in the CI environment. I'm not sure what's going on there, but the module isn't needed to build the docs. I've added the minimum modules required to the docs requirements file, and have redirected the workflows to install that instead of the module's full requirements file.

Also fixed a couple of build warnings while I was at it.


jlashner (approved):
Thanks for fixing this! Looks good

PR #320, opened 2022-12-15
bump dockerfile

jlashner:
Bump so_smurf_docker version. This fixes issues the error we were seeing before, and has been tested on the SAT1 system


msilvafe (approved):
discussed offline

PR #321, opened 2022-12-19
Make relock_tracking_setup use feedback start/end from cfg

jlashner:
Changes relock_tracking_setup so that it pulls feedback start frac and end frac from the device cfg.


msilvafe (approved):
This is a good fix, thanks!

PR #322, opened 2022-12-19
bugfix for bias_steps change

jlashner:
bugfix for bias step function


msilvafe (approved):
This is a good catch. Thanks for fixing.

PR #330, opened 2023-01-23
Adds downsample option to bias_steps

jlashner:
Also adds tau_eff fitting optional in bias step analysis.

Downsampling is useful for when we want to run bias steps on all slots simultaneously and don't care about tau_eff fitting.


jlashner:
Closing this PR, as getting the bias step analysis to work properly wasn't trivial with the lower sampling rate.

PR #331, opened 2023-01-24
Removes pysmurf type hints

jlashner:
I added these type hints since it is helpful for development + autocomplete, however it seems like they cause some import errors in certain environments like the pysmurf-controller, so probably best to remove.


jlashner:
Pysmurf controller is working well without this PR, so I am going to close this unless there is another reason to remove them.

PR #334, opened 2023-03-03
Don't assume bias steps start in low current mode

dpdutcher:
If high_current_mode=True, only modify the biases for the bias groups that were in low current mode to begin with.

### DIFF
--- a/sodetlib/operations/bias_steps.py
+++ b/sodetlib/operations/bias_steps.py
@@ -1020,11 +1020,12 @@ def take_bgmap(S, cfg, bgs=None, dc_voltage=0.3, step_voltage=0.01,
         nsteps (int):
             Number of steps to run
         high_current_mode (bool):
-            If true, switches to high-current-mode. If False, leaves in LCM
-            which runs through the bias-line filter, so make sure you
+            If True, switches bias groups to high-current-mode.
+            If False, switches bias groups to low-current-mode.
+            LCM runs through the bias-line filter, so make sure you
             extend the step duration to be like >2 sec or something
         hcm_wait_time (float):
-            Time to wait after switching to high-current-mode.
+            Time to wait after switching to/from high-current-mode.
         dacs : str
             Which group of DACs to play bias-steps on. Can be 'pos',
             'neg', or 'both'
@@ -1109,12 +1110,12 @@ def take_bias_steps(S, cfg, bgs=None, step_voltage=0.05, step_duration=0.05,
         nsteps (int):
             Number of steps to run
         high_current_mode (bool):
-            If true, switches to high-current-mode. If False, leaves in whichever
-            current mode was set initially.
+            If True, switches bias groups to high-current-mode.
+            If False, switches bias groups to low-current-mode.
             LCM runs through the bias-line filter, so make sure you
             extend the step duration to be like >2 sec or something
         hcm_wait_time (float):
-            Time to wait after switching to high-current-mode.
+            Time to wait after switching to/from high-current-mode.
         dacs : str
             Which group of DACs to play bias-steps on. Can be 'pos',
             'neg', or 'both'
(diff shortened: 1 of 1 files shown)


sodetlib/operations/bias_steps.py
```diff
@@ -1159,7 +1159,8 @@ def take_bias_steps(S, cfg, bgs=None, step_voltage=0.05, step_duration=0.05,
         dc_biases = initial_dc_biases
         init_current_mode = sdl.get_current_mode_array(S)
         if high_current_mode:
-            dc_biases = dc_biases / S.high_low_current_ratio
+            dc_biases[init_current_mode == 0] = (dc_biases[init_current_mode == 0] /
+                                                 S.high_low_current_ratio)
             step_voltage /= S.high_low_current_ratio
```
msilvafe, line 1164: @dpdutcher this `step_voltage` should probably also get the same treatment so that the step amplitude is equal for all bias lines independent of their current mode.

msilvafe (changes requested):
I think this fixes how the bias step is taken but I think the analysis function breaks. The computed parameters assume all bias lines are in lcm or hcm: 

https://github.com/simonsobs/sodetlib/blob/86d2e684efc76bdfacc2ec41f67cf95ee4ccf4bf/sodetlib/operations/bias_steps.py#L697

 but your logic here allows for the current mode to be different bias line by bias line (which I could imagine we want if the dichroic splits have very different PSATs or efficiencies). So I think before we merge this we want to fix the analysis functions to work with this change to the data taking.

sodetlib/operations/bias_steps.py
jlashner: I'm not sure if step_voltage needs to be changed, because the function parameter `step_voltage` should always be in low current mode since that is user controlled

msilvafe: Ahh ok @jlashner then do you think that the comment I made about the analysis function still holds then? Does bsa.high_current_mode need to be a boolean array instead of a boolean scalar?

jlashner: From what I can tell the analysis shouldn't need to be changed, but I think we should definitely test before merging. If I'm not mistaken all bias lines should still be in high current mode during the steps, so we should only need a single boolean. The change here is that if all bias lines are not in high current mode before the step, it will only correct the ones that are not.

jlashner:
Thanks Daniel, this looks good to me! I do want to make sure it's tested before merging so let me know if you've done that / when you are able to.

Though as this is now, having a combination of high-current-mode and low-current-mode bias lines will still break `take_bias_steps` when `high_current_mode=False` is passed. Is this something we're worried about? Or do we not care about this since for the high Psat detectors we'll never want to run bias steps in low-current-mode?

dpdutcher:
I tested this with Uv31 back when it was in Princeton and it worked--there I was doing bias_steps vs. Vbias and so needed to start bias steps off in high current mode. All bias lines were in high current mode, though, so I have not investigated what happens when operating with a mix of current modes. It's kind of a niche situation, but I think we should support it. Where precisely does it break?

sodetlib/operations/bias_steps.py
```diff
@@ -1159,7 +1159,8 @@ def take_bias_steps(S, cfg, bgs=None, step_voltage=0.05, step_duration=0.05,
         dc_biases = initial_dc_biases
         init_current_mode = sdl.get_current_mode_array(S)
         if high_current_mode:
-            dc_biases = dc_biases / S.high_low_current_ratio
+            dc_biases[init_current_mode == 0] = (dc_biases[init_current_mode == 0] /
```
jlashner, line 1162: This needs an else branch that will set the current-mode to 0 if we're assuming bias groups are already in low current mode.

dpdutcher:
Now I'm confused if this PR is even really needed. The initial_dc_biases and dc_biases variables weren't actually used anywhere, and those were the only significant things I changed in the original commit.

jlashner:
Ya I think those weren't actually used, but this PR does change the function so that it only runs `set_current_mode` on the

PR #336, opened 2023-03-07
Splits 'oper' tags from user-set g3 tags

jlashner:
For data packaging, we want to be able to pass tags in from the pysmurf-controller. As it is currently implemented, user-provided tags will overwrite the `oper` tags set by various sodetlib operations. To fix this, adds a separate `oper` keyword to the `stream_data_on` function, which is used to determine the initial part of g3 tag (i.e. `oper,iv` or `obs,stream`). The user-provided tag will be appended to the end.

This needs to be tested, preferably with the pysmurf-controller, before merging.

### DIFF
--- a/sodetlib/noise.py
+++ b/sodetlib/noise.py
@@ -510,7 +510,7 @@ def take_noise(S, cfg, acq_time=30, plot_band_summary=True, nbins=40,
         passing low-f requirement (i.e. fknee set by wl = 65pA/rtHz and
         slope must be <= 1/f^{1/2} in the ASD)
     g3_tag: string, optional
-        if not None, overrides default tag "oper,noise" sent to g3 file
+        Tag to be attached to g3 stream 
     
     Returns
     -------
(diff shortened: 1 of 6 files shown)


kmharrington:
This looks like it will work for operations, but how do we get tags from the observations that are `obs,cmb,<extra tags>` and `obs,cal,<extra_tags>`? The scheduler>>sorunlib>>controller needs to tell us which one we want but stream should just be the catch-all if nothing is specified.

jlashner:
Ok,  not sure if I understood the logic of the `obs` tags. Do we need the `stream` as a catch-all? If not we can just remove that and pass in `cmb,<extra-tags>` and `cal,<extra-tags>` through the `tag` param of the controller, and it'll just be `obs,<extra-tags>` as default.

I'd rather not have a soft requirement that the 2nd tag is the oper name if we can avoid it, since that is not documented anywhere and not really obvious... but if that's required then we can pass `oper=cmb/cal` and have it detect that and use `obs` instead of `oper`

kmharrington:
It is documented in the Flow to Librarian doc: https://simons1.princeton.edu/docs/tod2maps_docs/flow_to_librarian/  (section 5.1)

All tags are <type>,<subtype>,<extra tags>. There are subtypes of both observation and operation books. 

We do need the `stream` catch-all because both the `cmb` and `cal` tags should come from the scheduler / automated systems at the site. `stream` should be for when someone is just messing around at the site and starts data taking or something else like that. I don't want to use `cal` as a catch all because even those observations are "special" in that they're commanded through our automated systems.

jlashner:
Got it, thanks katie! I probably read that before and forgot about it.

So to enforce the tag structure I changed this slightly so that stream now accepts two arguments: `stream_type` (which can be 'obs' or 'oper'), and `subtype`, which is the second tag. By default, the stream_type will automatically be set to `obs` for "stream", "cmb", or "cal" subtypes, and will be set to `oper` otherwise. So the pysmurf controller will just need to pass `subtype='cmb'` + additional tags if we want. If you approve this I'll update the controller PR.

kmharrington (approved):
Tested here: https://github.com/simonsobs/socs/pull/424#pullrequestreview-1458009368

kmharrington (approved):
Tested here: https://github.com/simonsobs/socs/pull/424#pullrequestreview-1458009368

PR #337, opened 2023-03-15
Fixed get_wls_from_aman function to use new det_info convention inste…

msilvafe:
This was still using the old ch_info to get bands. I've updated it to use the new det_info structure.


jlashner:
Looks good to me, thanks max! Just out of curiosity, how was this causing a failure? I thought it would've been ok since we do this: https://github.com/simonsobs/sodetlib/blob/86d2e684efc76bdfacc2ec41f67cf95ee4ccf4bf/sodetlib/stream.py#L101-L102

jlashner (approved):

msilvafe:
Jack's comment made me realize I was using some outdated sodetlib branch on this system. This is fixed by the lines Jack pointed out so I'll close this.

PR #338, opened 2023-03-15
small bias step fixes

jlashner:
Small bugfixes to the bias_steps module
 - Fix plot_steps function
 - Cut channels based on `np.abs(R0)` instead of `R0`


msilvafe (approved):
Thanks for fixing these.

PR #339, opened 2023-03-21
Adds fixed_tone_loopback test to QC

jlashner:
QC function that enables many (2048) fixed tones and measures noise across the band. Creates plots like this:
image

Takes roughly 2 min to run for 3 measured freqs per band, and ~10 min for 10 measured chans per band.

### DIFF
--- a/sodetlib/quality_control.py
+++ b/sodetlib/quality_control.py
@@ -1,7 +1,10 @@
 import sodetlib as sdl
 import numpy as np
 import time
+from scipy import signal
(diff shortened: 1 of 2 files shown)


msilvafe (approved):
Make the couple changes then good to merge.

jlashner:
Thanks max! Just submitted a PR that addresses your comments.

msilvafe:
Great, thanks. I will merge then.

PR #340, opened 2023-03-30
Add check for which version of bias step loader we're running

kmharrington:
We changed how sotodlib loads smurf channel information. This let's the bias step analysis loader work on both old and new versions.


jlashner (approved):
Looks good to me! We should probably begin to deprecate using `am.ch_info` and switch everything to `det_info.smurf`

PR #341, opened 2023-04-12
load data needs stream_id too

kmharrington:
Error found while running some reanalysis off the smurf server.


msilvafe (approved):
Good catch Katie.

PR #344, opened 2023-05-18
Added ability for synthesis_scale to be set per band in dev_cfg

msilvafe:
For @johncgroh to be able to adjust synthesis scale per band in dev_cfg when using high band AMC that requires lower synthesis scale.


jlashner (approved):
Looks good to me! If you get to test this feel free to merge.

PR #362, opened 2023-06-20
Biasstep rebias

yuhanwyhan:
this contains the script of re-biasing detectors using biasstep.
this function has recently been tested in LATR-t with different loading conditions.
this plot is showing the result of re-biasing detectors when LATR-t ful exposed to the room
Screenshot 2023-06-20 at 11 58 53 AM
this plot is showing the result of re-biasing detectors after covering the LATR-t with a metal plate
Screenshot 2023-06-20 at 11 59 40 AM
Notice the tests were done with UHF detectors (Uv31), and the change of loading between these two test are very large.
in the field we expect lower and more gradually change of loadings.

The next level thing is to make the script run faster, this might involve only running part of the analysis code for bias-step.

### DIFF
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -424,7 +424,7 @@ def biasstep_rebias(
 
     S.log("applying the new bias voltages")
     S.set_tes_bias_bipolar_array(vbias_estimate)  
-    S.log(f"applying {set_tes_bias_bipolar_array}")
+    S.log(f"applying {vbias_estimate}")
     S.log("taking the final biasstep")
     bsa_2 = bias_steps.take_bias_steps(S, cfg)
     if not math_only:
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -3,7 +3,7 @@ import sodetlib as sdl
 import time
 import matplotlib.pyplot as plt
 from sodetlib.operations.iv import IVAnalysis
-from sodetlib.operations import uxm_setup, uxm_relock, tracking, bias_steps, iv, bias_dets
+from sodetlib.operations import bias_steps, iv
 np.seterr(all="ignore")
 
 
@@ -203,23 +203,52 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
 
 def biasstep_rebias(
         S, cfg, target_percentage_rn = 0.5, bias_groups=None,
-        math_only=False):
+        show_plots=True, make_plots=True):
 
     """
-    Scripts work well when detectors are already in transistion. For extreme cases
-    such as detectors are all in SC, it takes a few run to put them perfectly in
-    transition. Next step is making this script can handle extreme cases faster
+    re-bias detectors by taking bias_steps.
+    
+     Args
+    -----
+    S : SmurfControl
+        Pysmurf instance
+    cfg : DetConfig
+        det config instance
+    target_percentage_rn : float
+        this is the target percentage rn for the script to bias to, numbers should be
+        between 0.3 and 0.7 for general usage.
+    bias_groups : list
+        List of bias groups to bias. If None, this will default to the active
+        bgs specified in the device cfg.
+    make_plots : bool
+        If this is set to False, the script will still apply the bias-voltage but will not
+        generate plots
+    show_plots : bool
+        If this is set to False, the script will not show plots
 
-
+    Returns
+    ----------
+    bsa_final :
+        analysis result of the last biasstep taken with the re-biased bais voltage
+    vbias_estimate_final :
+        final bias_voltages calculated and applied by the script
     """
+    if bias_groups is None:
+        bias_groups = cfg.dev.exp['active_bgs']
+    bias_groups = np.atleast_1d(bias_groups)
 
     ## take the initial biasstep
     S.log("taking the first biasstep")
-    intical_current_biasvoltage = S.get_tes_bias_bipolar_array()
-    S.log(f"current biasvoltage {intical_current_biasvoltage}")
+    S.log(f"Initial dc biases {S.get_tes_bipolar_array()}")
     bsa_0 = bias_steps.take_bias_steps(S, cfg)
-    if not math_only:
-        bias_steps.plot_Rfrac(bsa_0)
+    if make_plots:
+        fig, ax = bias_steps.plot_Rfrac(bsa_0)
+        fname = sdl.make_filename(S, 'intial_Rfrac.png', plot=True)
+        fig.savefig(fname)
+        S.pub.register_file(fname, 'biasstep_rfrac', format='png', plot=True)
+        if not show_plots:
+            plt.close(fig)
+            
             
     ## load IV analysis result so can use dynamic step size
     iva = iv.IVAnalysis.load(bsa_0.meta['iv_file'])
@@ -374,8 +403,13 @@ def biasstep_rebias(
     ## taking the second bias step
     S.log("taking the 2nd biasstep")
     bsa_1 = bias_steps.take_bias_steps(S, cfg)
-    if not math_only:
-        bias_steps.plot_Rfrac(bsa_1)
+    if make_plots:
+        fig, ax = bias_steps.plot_Rfrac(bsa_1)
+        fname = sdl.make_filename(S, '2nd_Rfrac.png', plot=True)
+        fig.savefig(fname)
+        S.pub.register_file(fname, 'biasstep_rfrac', format='png', plot=True)
+        if not show_plots:
+            plt.close(fig)
 
     ## check for script failue
     for bl in bias_groups:
@@ -435,8 +469,13 @@ def biasstep_rebias(
         bsa_2 = bias_steps.take_bias_steps(S, cfg)
         repeat_biasstep = False
 
-    if not math_only:
-        bias_steps.plot_Rfrac(bsa_2)
+    if make_plots:
+        fig, ax = bias_steps.plot_Rfrac(bsa_2)
+        fname = sdl.make_filename(S, 'rebiased_Rfrac.png', plot=True)
+        fig.savefig(fname)
+        S.pub.register_file(fname, 'biasstep_rfrac', format='png', plot=True)
+        if not show_plots:
+            plt.close(fig)
 
 
     S.log("confirming if the current result is close to the target")
@@ -498,8 +537,13 @@ def biasstep_rebias(
             bsa_extra = bias_steps.take_bias_steps(S, cfg)
             repeat_biasstep = False
         
-        if not math_only:
-            bias_steps.plot_Rfrac(bsa_extra)
+        if make_plots:
+            fig, ax = bias_steps.plot_Rfrac(bsa_extra)
+            fname = sdl.make_filename(S, '3rd_Rfrac.png', plot=True)
+            fig.savefig(fname)
+            S.pub.register_file(fname, 'biasstep_rfrac', format='png', plot=True)
+            if not show_plots:
+                plt.close(fig)
 
         percentage_rn_extra = bsa_extra.R0/bsa_extra.R_n_IV
 
@@ -532,8 +576,13 @@ def biasstep_rebias(
         S.log(f"applying {vbias_estimate_final}")
         S.log("taking the final biasstep")
         bsa_final = bias_steps.take_bias_steps(S, cfg)
-        if not math_only:
-            bias_steps.plot_Rfrac(bsa_final)
+        if make_plots:
+            fig, ax = bias_steps.plot_Rfrac(bsa_final)
+            fname = sdl.make_filename(S, 'rebiased_Rfrac_2.png', plot=True)
+            fig.savefig(fname)
+            S.pub.register_file(fname, 'biasstep_rfrac', format='png', plot=True)
+            if not show_plots:
+                plt.close(fig)
 
 
     return bsa_final,vbias_estimate_final
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -239,7 +239,7 @@ def biasstep_rebias(
 
     ## take the initial biasstep
     S.log("taking the first biasstep")
-    S.log(f"Initial dc biases {S.get_tes_bipolar_array()}")
+    S.log(f"Initial dc biases {S.get_tes_bias_bipolar_array()}")
     bsa_0 = bias_steps.take_bias_steps(S, cfg)
     if make_plots:
         fig, ax = bias_steps.plot_Rfrac(bsa_0)
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -459,7 +459,7 @@ def biasstep_rebias(
         per_bl_vbias_estimate = vbias_estimate_array[mask_bg]
         med_per_bl_vbias_estimate = np.nanmedian(per_bl_vbias_estimate)
         if np.isnan(med_per_bl_vbias_estimate):
-            med_per_bl_vbias_estimate = (v1 + v0)/2
+            med_per_bl_vbias_estimate = (np.nanmedian(v1[mask_bg]) + np.nanmedian(v0[mask_bg]))/2
         vbias_estimate[bl] = med_per_bl_vbias_estimate
 
     S.log("applying the new bias voltages")
@@ -579,7 +579,7 @@ def biasstep_rebias(
             per_bl_vbias_estimate = vbias_estimate_array_extra[mask_bg]
             med_per_bl_vbias_estimate = np.nanmedian(per_bl_vbias_estimate)
             if np.isnan(med_per_bl_vbias_estimate):
-                med_per_bl_vbias_estimate = (v1 + v0)/2
+                med_per_bl_vbias_estimate = (np.nanmedian(v1[mask_bg]) + np.nanmedian(v0[mask_bg]))/2
             vbias_estimate_final[bl] = med_per_bl_vbias_estimate
 
 
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -266,12 +266,12 @@ def biasstep_rebias(
 
 
     if overbias_needed:
-        S.log("some BL are stuck in SC, going to over bias them, this takes about 4mins")
+        S.log("some BL are stuck in SC, going to over bias them, this takes about 5mins")
         previous_dc_biases = S.get_tes_bias_bipolar_array()
         
         sdl.overbias_dets(S, cfg, bias_groups=bg_overbias_needed)
-        ##force wait time 3 mins for UFM to reach equalibrium 
-        for sleep_index in range(3):
+        ##force wait time 5 mins for UFM to reach equalibrium 
+        for sleep_index in range(5):
             time.sleep(60)
         safe_dc_biases = previous_dc_biases.copy()
         for replace_bg in bg_overbias_needed:
@@ -325,9 +325,10 @@ def biasstep_rebias(
             new_bias_voltage[bl] =  initial_dc_biases[bl] - delta_dc_bias
         ## if the detector is normal, take a larger step
         if bl in bg_detectors_normal:
+            S.log(f"some detectors are in normal state for bl{bl}, taking a large step")
             if initial_dc_biases[bl] > cfg.dev.bias_groups[bl]["testbed_100mK_bias_voltage"]:
                 initial_dc_biases[bl] = cfg.dev.bias_groups[bl]["testbed_100mK_bias_voltage"]
-            new_bias_voltage[bl] =  initial_dc_biases[bl] - 0.5 * v_spread
+            new_bias_voltage[bl] =  initial_dc_biases[bl] - 0.3 * v_spread
 
     S.log(f"applying new voltage for 2nd biasstep {new_bias_voltage}")
     S.set_tes_bias_bipolar_array(new_bias_voltage)  
@@ -382,6 +383,8 @@ def biasstep_rebias(
         
         S.set_tes_bias_bipolar_array(previous_dc_biases_dfn)
         S.log(f"applying: {previous_dc_biases_dfn}")
+        S.log(f"waiting 30s for dissipation of UFM local heating")
+        time.sleep(30)
         bsa_1 = bias_steps.take_bias_steps(S, cfg)
         percentage_rn_1 = bsa_1.R0/bsa_1.R_n_IV
         drop_from_normal = False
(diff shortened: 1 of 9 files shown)


sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+            overbias_needed = True
+
+    ## add in check if detectors are normal, hence we should increase the drop step
+    bg_detectors_normal = []
+    drop_from_normal = False
+    for bl in bias_groups:
+        mask_bg = np.where(bsa_0.bgmap==bl)[0]
+        per_bl_percentage_rn_0 = percentage_rn_0[mask_bg]
+        ## mask for normal
+        mask_normal = np.where((per_bl_percentage_rn_0 > 0.9) | (per_bl_percentage_rn_0 < -0.9))
```
dpdutcher, line 263: np.abs?

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+        S.log(f"some BL: {bg_overbias_needed} are stuck in SC, going to over bias them, this takes about 5mins")
+        previous_dc_biases = S.get_tes_bias_bipolar_array()
+        
+        sdl.overbias_dets(S, cfg, bias_groups=bg_overbias_needed)
+        ##force wait time 5 mins for UFM to reach equalibrium 
+        for sleep_index in range(5):
+            time.sleep(60)
+        safe_dc_biases = previous_dc_biases.copy()
+        for replace_bg in bg_overbias_needed:
+            safe_dc_biases[replace_bg] = cfg.dev.bias_groups[replace_bg]["testbed_100mK_bias_voltage"]
```
dpdutcher, line 280: Is that a standard entry in the config file? In other words, is it guaranteed to exist?

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
 
     return biases
 
+def biasstep_rebias(
+        S, cfg, target_percentage_rn = 0.5, bias_groups=None,
+        math_only=False):
```
dpdutcher, line 206: Plotting should be optional, and plots should be published.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+        drop_from_normal = False
+        for bl in bg_overbias_needed:
+            mask_bg = np.where(bsa_0.bgmap==bl)[0]
+            per_bl_percentage_rn_0 = percentage_rn_0[mask_bg]
+            ## mask for normal
+            mask_normal = np.where((per_bl_percentage_rn_0 > 0.9) | (per_bl_percentage_rn_0 < -0.9))
+            ## if more than half of the detectors are normal
+            if len(mask_normal[0]) > 0.5*len(mask_bg):
+                bg_detectors_normal.append(bl)   
+                drop_from_normal = True
```
dpdutcher, line 297: This code block is repeated from before (I think). Could be tidied up to avoid repetition.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
 
+def biasstep_rebias(
+        S, cfg, target_percentage_rn = 0.5, bias_groups=None,
+        math_only=False):
+
+    """
+    Scripts work well when detectors are already in transistion. For extreme cases
+    such as detectors are all in SC, it takes a few run to put them perfectly in
+    transition. Next step is making this script can handle extreme cases faster
+
```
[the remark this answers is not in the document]
dpdutcher: Also an overview of how the script works would be nice.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+    if repeat_biasstep:
+        bsa_1 = bias_steps.take_bias_steps(S, cfg)
+        repeat_biasstep = False
+
+    percentage_rn_1 = bsa_1.R0/bsa_1.R_n_IV
+
+    target_percentage_rn_array = target_percentage_rn * np.ones(len(percentage_rn_0))
+    ## getting result from the previous 2
+    v0 = np.zeros(bsa_0.bgmap.shape[0])
+    for bl in range(12):
```
dpdutcher, line 398: Here and in line 403, use `bias_groups` and not `range(12)`

sodetlib/operations/bias_steps.py
```diff
@@ -960,7 +960,7 @@ def plot_Rfrac(bsa, text_loc=(0.6, 0.8)):
     chmap = sdl.map_band_chans(bsa.bands, bsa.channels,
                                iva.bands, iva.channels)
 
-    Rfracs = bsa.R0 / iva.R_n_IV[chmap]
+    Rfracs = bsa.R0 / iva.R_n[chmap]
```
dpdutcher, line 963: This change was already made in master, so heads up that you will probably need to manually merge some files when committing.

dpdutcher (changes requested):
From what I understand, the functionality is good, so my comments are more code hygiene things. There's some excess whitespace, long lines, and repeated sections of code that should be cleaned up. Doing so will help with readability and make debugging easier if needed.

sodetlib/operations/bias_dets.py
```diff
@@ -3,7 +3,7 @@
 import time
 import matplotlib.pyplot as plt
 from sodetlib.operations.iv import IVAnalysis
-
+from sodetlib.operations import uxm_setup, uxm_relock, tracking, bias_steps, iv, bias_dets
```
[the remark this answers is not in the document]
yuhanwyhan: fixed

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,347 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+    such as detectors are all in SC, it takes a few run to put them perfectly in
+    transition. Next step is making this script can handle extreme cases faster
+
+
+    """
+
+    ## take the initial biasstep
+    S.log("taking the first biasstep")
+    intical_current_biasvoltage = S.get_tes_bias_bipolar_array()
+    S.log(f"current biasvoltage {intical_current_biasvoltage}")
```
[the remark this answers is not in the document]
yuhanwyhan: fixed

sodetlib/operations/bias_dets.py
yuhanwyhan: I decided to keep it this way for now, this is for catching a bias-step failure as the current bias_step can return negative bias_resistance. The -0.9 threshold might be changed later with better understanding of the failure mode

sodetlib/operations/bias_dets.py
yuhanwyhan: this is not a standard entry but should be added in.
the purpose for this entry is so the script has a reference of what bias_voltage to use such that the detectors won't be in SC state at the site. This can also be understood as the bias_voltages at the minimum loading.
since all modules are tested in Princeton first, the dark detector bias_voltages are naturally the bias_voltages at the minimum loading. If this info is missing for some reason, user can put in some high numbers such as 15V for Uv31 as a start point.

sodetlib/operations/bias_dets.py
yuhanwyhan: fixed

sodetlib/operations/bias_dets.py
yuhanwyhan: This code block is not repeated. this code block is designed to be ran again if over_biasing is needed. After the overbiasing, the script has to choose some bias_voltage to land on, in order to be safe from dropping to SC again, the script will choose the biasvoltage recorded in the config file as "testbed_100mK_bias_voltage", this however might land the detectors in normal state. Hence this code block is for catching this after the overbiasing.

sodetlib/operations/bias_dets.py
yuhanwyhan: range(12) is used such that the bias voltages in range(12) but not in bias_groups can be kept even that bias_line was not chosen to be re-biased. The 12 is the upper bar of how many biaslines SO UFMs can have.

sodetlib/operations/bias_steps.py
yuhanwyhan: will do

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,396 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+        ##force wait time 5 mins for UFM to reach equalibrium 
+        for sleep_index in range(5):
+            time.sleep(60)
+        safe_dc_biases = previous_dc_biases.copy()
+        for replace_bg in bg_overbias_needed:
+            safe_dc_biases[replace_bg] = cfg.dev.bias_groups[replace_bg]["testbed_100mK_bias_voltage"]
+            
+        S.set_tes_bias_bipolar_array(safe_dc_biases)  
+        bsa_0 = bias_steps.take_bias_steps(S, cfg)
+        bias_steps.plot_Rfrac(bsa_0)
```
dpdutcher, line 313: Also need some if statements here about plotting

jlashner:
Hi Yuhan, I'm about to start reviewing this, but do you think you can post here (or in the docstring) like the overall strategy for how this works? Something like a flow-chart that documents what it will to do at each step under what conditions, etc. would be really helpful for understanding this at a higher level

sodetlib/operations/bias_dets.py
yuhanwyhan: ah, right, fixed

yuhanwyhan:
> Hi Yuhan, I'm about to start reviewing this, but do you think you can post here (or in the docstring) like the overall strategy for how this works? Something like a flow-chart that documents what it will try to do at each step under what conditions, etc. would be really helpful for understanding this at a higher level

the main logic is described in 4.2.2 of
https://arxiv.org/pdf/2208.05997.pdf

extra parts of this script is for handling failures of bias_step and handling extreme situations of detectors stucked in SC state or in normal state

sodetlib/operations/bias_dets.py
dpdutcher: Can you do some combination of (a) adding that parameter to [det_config.py](https://github.com/simonsobs/sodetlib/blob/master/sodetlib/det_config.py) (b) using a `.get()` call here to return a sensible default if no matching entry is found?

sodetlib/operations/bias_dets.py
dpdutcher: What I'm trying to point out is that https://github.com/simonsobs/sodetlib/blob/05fea5d1b4fcdea22c099d60ce505abc5ab64626/sodetlib/operations/bias_dets.py#L325-L332 are the exact same lines as https://github.com/simonsobs/sodetlib/blob/05fea5d1b4fcdea22c099d60ce505abc5ab64626/sodetlib/operations/bias_dets.py#L289-L296

so it could be, for example, condensed into a helper function

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+        per_bl_R0 = bsa_0.R0[mask_bg]
+        if np.isnan(np.nanmedian(per_bl_R0)):
+            ## the yield is very low, could it be the biasstep analysis script failed? repeat again
+            repeat_biasstep = True
+            S.log("the biasstep script might have failed, going to repeat the measurement")
+
+    if repeat_biasstep:
+        S.log("retaking biasstep due to failure")
+        bsa_0 = bias_steps.take_bias_steps(S, cfg)
+        repeat_biasstep = False
```
jlashner, line 269: I don't really like that this is forcing a retake of the bias steps, especially since if it fails a second time it will just continue on as if everything's ok. I think instead if the first bias steps fails, this should raise an exception with a useful error message.

Do you know how often this fails / why? We should try to fix in the bias step function directly instead of building fixes into this function.

yuhanwyhan: bias step failure has been noticed in multiple places including at the site in LATR, in Chicago LATRt, TSAT and Cornell. One of the failure mode might due to the if statement here: https://github.com/simonsobs/sodetlib/blob/4aafb94bb36cac85b754b0a1e28575d33e1b6243/sodetlib/operations/bias_steps.py#L645
Tho repeating the biasstep is not ideal, it is a working method given the current circumstances.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+
+
+    if overbias_needed:
+        S.log(f"some BL: {bg_overbias_needed} are stuck in SC, going to over bias them, this takes about 5mins")
+        previous_dc_biases = S.get_tes_bias_bipolar_array()
+        
+        sdl.overbias_dets(S, cfg, bias_groups=bg_overbias_needed)
+        ##force wait time 5 mins for UFM to reach equalibrium 
+        for sleep_index in range(5):
+            time.sleep(60)
```
jlashner, line 306: I think the sleep-time after overbiasing should be a configurable parameter, ideally settable in the dev-cfg

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+        safe_dc_biases = previous_dc_biases.copy()
+        for replace_bg in bg_overbias_needed:
+            try:
+                safe_dc_biases[replace_bg] = cfg.dev.bias_groups[replace_bg]["testbed_100mK_bias_voltage"]
+            except KeyError:
+                mask_bg = np.where(bsa_0.bgmap==replace_bg)[0]
+                v_norm_bl = np.nanmedian(iva.v_bias[iva.idxs[[mask_bg], 1]])
+                safe_dc_biases[replace_bg] = v_norm_bl
+
```
jlashner, line 315: Don't need this except anymore since it's in the default dev cfg dict.

sodetlib/operations/bias_dets.py
yuhanwyhan: but simply repeating it worked. yes Yaqiong and I have been reviewing some functions together.In my opinion I think the if statement can just be removed as we don't really rely on a accurate tes resistance from the biasstep when TES is not in-transition.

sodetlib/operations/bias_dets.py
yuhanwyhan: is there a way in the current det_config.py such that we can load different numbers for 90s and 150s?

sodetlib/operations/bias_dets.py
yuhanwyhan: yeah good idea.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+            mask_normal = np.where((per_bl_percentage_rn_0 > 0.9) | (per_bl_percentage_rn_0 < -0.9))
+            ## if more than half of the detectors are normal
+            if len(mask_normal[0]) > 0.5*len(mask_bg):
+                drop_from_normal = True
+            else:
+                bg_detectors_normal.remove(bl)
+        for bl in bg_detectors_normal:
+            if previous_dc_biases_dfn[bl] < 0.5:
+                bg_detectors_normal.remove(bl)
+
```
jlashner, line 380: Can you explain a bit about what this is doing? I don't quite understand how you're chosing the bias voltage step sizes, and also it's not clear to me from the code how many times this runs on average, but seems like it could be a lot of bias step measurements. Do you think this is mainly what's dominating the function's runtime?

jlashner: I think we might want to brainstorm other ways to drop into the transition. I think if we're taking small enough step sizes we may not need to take bias steps every time, since we're unlikely to induce tracking skips.

sodetlib/operations/bias_dets.py
jlashner: Do you think it's mostly failing when a large fraction of detectors are superconducting? I notice you don't have retries for other times you call bias steps.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
...
+            else:
+                previous_dc_biases_dfn[bl] =  previous_dc_biases_dfn[bl] - 0.3 * v_spread 
+                bg_detectors_normal.remove(bl)
+                S.log(f'biasline {bl} is approching 0 voltage, a confirmation new IV curve is recommended')
+        
+        S.set_tes_bias_bipolar_array(previous_dc_biases_dfn)
+        S.log(f"applying: {previous_dc_biases_dfn}")
+        S.log(f"waiting 10s for dissipation of UFM local heating")
+        time.sleep(10)
+        bsa_0 = bias_steps.take_bias_steps(S, cfg)
```
jlashner, line 364: If you do not need to run on all bias lines, you can specify the `bgs` arg here to just run on a selection of them. This may speed up the function.

sodetlib/operations/bias_dets.py
yuhanwyhan: this is looping down the bias voltages until a certain bias line is no longer in normal state and TES start to be in-transtion. The step size being the 50% of the transition region from the latest IV such that the bias_line won't be dropped into SC due to step being too large or the script takes too long due to the step being to small. Notice the latest IV doesn't necessary provides good estimation as the I-V won't be taken as frequently as the bias-step, so a larger step size here can be dangerous. 
in the field I think the function runtime is dominated by the bias-step analysis itself. This is because in the field the sky loading changes gradually and slowly. the longest runtime will be the wait time of biasing detectors out of SC state, but this shouldn't happen if the re-biasing is done frequently enough nor should dropping from normal state happen frequently.

sodetlib/operations/bias_dets.py
yuhanwyhan: this is a good idea, but I think this should be added into the next generation of this code. This is because I currently rely on bg map for all 12 biaslines in this version of code.

sodetlib/operations/bias_dets.py
yuhanwyhan: i agree.
speaking of this, what should be the correct way of generating a config file. I am not sure generating from the default dev cfg dict is practiced a lot currently in different places

jlashner: In your sysconfig, if you set the devcfg path as one that doesn't exist, it will try to create a new one there with the defaults

sodetlib/operations/bias_dets.py
yuhanwyhan: I added in a retry for each bias step I think. I think the failing can happen either when a large fraction of detectors are superconducting or being normal. 
also repeating biasstep twice is like a old voodoo from my ACT days. I think such fix might make sense if one imagines some ADC needs to be 'waken'. I am not sure but I agree this can be redundant and can be removed after better understanding later.

sodetlib/operations/bias_dets.py
yuhanwyhan: oh yeah yeah good point... but I am hoping it is ok to just leave it like this as it just used only twice

sodetlib/operations/bias_dets.py
yuhanwyhan: Screenshot 2023-06-20 at 6 57 13 PM
Screenshot 2023-06-20 at 6 57 38 PM
this is one of the failure case I just found
after taking another bias_step this is fixed 
this is taken in LATRt, on Uv31, I believe the session id is 1687298049

yuhanwyhan: failure:
Screenshot 2023-06-20 at 7 00 48 PM
279dfe8-53f4-48de-b175-ec6991953072">
d19-1483-4a48-8419-e01e49fb9f56">
after repeating
Screenshot 2023-06-20 at 7 01 39 PM
Screenshot 2023-06-20 at 7 01 52 PM
Screenshot 2023-06-20 at 7 02 01 PM

jlashner: awesome, that's really interesting. are these files all on simons1? I'll take a look tomorrow.

jlashner:
Hi Yuhan, finally was able to read through the entire function. Here are the notes I took about the control flow. Do you agree with this  as the intended behavior?

1. Take initial bias steps `bsa_0`
    - will rebias if there are any bias lines where R0 is all nan
    - If >10% of dets are have Rfrac < 0.1, marks bg as overbias_needed
   - If >50% of dets have Rfrac > 0.9, marks bg as normal (and drop from normal needed)
2. Overbias `overbias_needed` bgs
    - runs `overbias_dets` on bgs that need overbiasing, and waits
    - For biaslines that have been overbiased, set the DC bias voltage to `testbed_100mK_bias_voltage`
    - Run bias steps
        - (no bias retry here)
    - Check normal condition (above) and add updated bias lines to `bg_detectors_normal` array if they meet it
3. Drop from normal
    - For each normal bg, step voltage by 0.5(V_norm - V_SC) determined from IV
    - Run bias steps 
         - update bsa_0 so it's always the last bias step ran
         - These bias steps don't have a retry
    - Check normal condition. If any bgs are still normal, repeat.
    - After exiting loop, the resulting DC biases are `initial_dc_biases`
4. Find new bjas voltage and take bias steps:
    - `vspread` is diff between median normal and median SC voltages based on IV
    - New bias voltage is `init +/- 0.15*vspread` where `+/-` is determined by
      whether or not the median Rfrac is > or < than 0.5
    - Take bias steps (`bsa_1`) and retry if needed
5. Determine new bias point for each channel:
    - `v_estimate = (v0 (target - Rfrac_1) + v1 (Rfrac_0 - target)) / (Rfrac_0 - Rfrac_1)`
    - `vnew = median(v_estimate[bg])`, or if all are nan, `vnew = (v1 - v0) / 2`
6. Set bias to `vnew` and retake bias steps `bsa_2` (with retry if needed)
    - BG is successful if `abs(Rfrac - target) < 0.05`
7. Fine tuning unsuccessful bias groups
    - Step 0.15 `vspread` in the appropriate direction
    - Take bias steps (with retry)
    - `v_estimate2 = (v0 (target - Rfrac_1) + v1 (Rfrac_0 - target)) / (Rfrac_0 - Rfrac_1)`
        - with `v0= prev estimate` and `v1 = prev estimate + delta`
    - Apply `v_estimate2`, retake bias steps.
8. Return  final bias step and bias voltages

jlashner:
Can you give an estimate of the run-time?

yuhanwyhan:
> Can you give an estimate of the run-time?

4 mins if start from in-transition, 14 mins if start from SC. (tested in LATRt)

msilvafe:
I like this description of the flow of the function. Since @yuhanwyhan gave the thumbs up to the description can you commit that to the docs rst file along with a link to Yuhan's conference proceedings @jlashner ?

yuhanwyhan:
there is one potential rare failure mode I just realized.
let's assume a situation that there is one bad biasline that causes heating or that biasline is short to a FR, and we don't want to touch it at all. The current script will still take a few bias_steps on this biasline. Although if a biasline is bad, we will not put bias_voltage on it to start with, so taking a biasstep round 0 voltage shouldn't do much damage, still....

the solution for this issue will require some re-writing.

the current script forces taking bg map for all 12 biaslines because when it calculates the v_estimate, it does the calculation for all biaslines and then apply the bg_map. One possible fix will be adding a for loop looping over the chosen biaslines and apply the bg_map inside the for loop such that the script only calculate the v_estimate for the chosen biaslines.

I suggest adding this fix into the later version of this script when we can test it cold again. Right now without a cold cryostat and MF/UHF UFMs, I am not confident that this change won't break the code.

yuhanwyhan:
a way to make the script faster will be only running the R_tes calculation from biasstep except the final biasstep (running only the _compute_R0_I0_Pj part). An estimation of the run-time for each analysis block of the current biasstep function will be nice.

dpdutcher (approved):

jlashner (changes requested):
Since this has been tested, I'm fine with merging this now with minimal changes to the code. I think we can merge as long as we make the following updates to the documentation:
- Add a link to the spie proceeding describing the procedure in the docstring
- Add my step-by-step overview of what the function is doing to the docstring
- Add a comment or some sort of documentation describing how `testbed_100mK_bias_voltage` is being used. I'm not exactly sure how this is used right now, but think we may want to change the name a bit, since this will be used in non-testbed systems, and I think this should probably have the word "transition" in it somewhere. Maybe `100mK_transition_bias_voltage`? 

As we test this more and start to use it in the field, I think we will probably want to clean up / update the functionality, but that can all go into separate PRs.

sodetlib/operations/bias_dets.py
kmharrington: Could we name this something other than testbed bias voltage? I saw "testbed" and was very confused about why that was useful on-site until I saw you explanation here. 

"dark_bias_voltage" or even "reference_bias_voltage" could make this clearer. And it would (correctly) imply that we could change these numbers later as we start to know the UFMs and on-sky loading better.

sodetlib/operations/bias_dets.py
```diff
@@ -201,3 +201,408 @@ def bias_to_rfrac(S, cfg, rfrac=0.5, bias_groups=None, iva=None,
 
     return biases
 
+def biasstep_rebias(
```
kmharrington, line 204: Could we set this function as an action? We will want this to be categorized as a single detector operation instead of "many many" bias steps in the data packaging. We'll also want to change the default tags getting added to these .g3 files.

jlashner (approved):
Added a simple docs page linking to the spie paper, and made the function a pysmurf action. I think I am happy to merge this in now and solve any problems with it as they come up.

msilvafe:
> Added a simple docs page linking to the spie paper, and made the function a pysmurf action. I think I am happy to merge this in now and solve any problems with it as they come up.

@kmharrington mentioned 2 additional things needed for this to be used within the site data operations, and packaging:

1. Getting the .g3 tags in the file to be something other than oper,bias_steps so they aren't all made into different operations books.
2. Figure out how to get imprinter to bind them all together into one book

I’m happy to wait on 2 until later but if 1 is a simple-ish modification I would like to get that in before merge. We want to use this soon in SAT-MF1 though so if not simple ok to merge and open new issue/or with these additional things.

jlashner:
Right, thanks for reminding me. Latest commit should do that, and sets all g3_tags to `biasstep_rebias`

msilvafe:
> Right, thanks for reminding me. Latest commit should do that, and sets all g3_tags to `oper,biasstep_rebias`

Thanks let's get this merged then

Issue #214, opened 2022-02-22
load_session call in bias step analysis does not respect g3_dir from cfg


dpdutcher:
https://github.com/simonsobs/sodetlib/blob/061f4635b09e6856ad90cac2331512cc7d029ec0/sodetlib/operations/bias_steps.py#L315

PR #365, opened 2023-06-29
Add  `base_dir` kwarg to bias_steps.run_analysis

dpdutcher:
Addresses #214 

Tested at Princeton.

### DIFF
--- a/sodetlib/operations/bias_steps.py
+++ b/sodetlib/operations/bias_steps.py
@@ -359,8 +359,8 @@ class BiasStepAnalysis:
 
     def run_analysis(
             self, create_bg_map=False, assignment_thresh=0.3, save_bg_map=True,
-            arc=None, step_window=0.03, fit_tmin=1.5e-3, transition=None,
-            R0_thresh=30e-3, save=False, bg_map_file=None):
+            arc=None, base_dir='/data/so/timestreams', step_window=0.03, fit_tmin=1.5e-3,
+            transition=None, R0_thresh=30e-3, save=False, bg_map_file=None):
         """
         Runs the bias step analysis.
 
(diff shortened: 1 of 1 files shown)


sodetlib/operations/bias_steps.py
```diff
@@ -467,7 +470,8 @@ def _load_am(self, arc=None, fix_timestamps=True):
             if arc:
                 self.am = arc.load_data(self.start, self.stop)
             else:
-                self.am = sdl.load_session(self.meta['stream_id'], self.sid)
+                self.am = sdl.load_session(self.meta['stream_id'], self.sid,
```
msilvafe, line 473: It doesn't look like `load_session` can currently take a `base_dir` argument. Also, out of curiosity, what issue is this PR addressing?

dpdutcher: https://github.com/simonsobs/sodetlib/blob/4aafb94bb36cac85b754b0a1e28575d33e1b6243/sodetlib/stream.py#L71-L72

msilvafe: Interesting...I was looking at `load_session` here: https://github.com/simonsobs/sodetlib/blob/4aafb94bb36cac85b754b0a1e28575d33e1b6243/sodetlib/legacy/smurf_funcs/smurf_ops.py#L422 not sure why we have a duplicate.

jlashner (approved):
Looks good to me! I don't love hardcoding `/data/so/timestreams` but I take full responsibility for that bc how  `load_session` is implemented, and recognize there is not really an easy robust way to default to `cfg.g3_dir`

sodetlib/operations/bias_steps.py
msilvafe: Ahh it's in "legacy" just searched in github and it gave me that one. Ok, this all makes sense we should just accept this PR probably but I'll wait an hr or so for @jlashner to have a chance to take a look.

dpdutcher: I linked Issue #214 , which has a slightly different title, but the key aspect there being there was no base_dir kwarg being passed to the load_session call in the bias step analysis.

This problem arises when trying to analyze bias steps without an G3tSMuRF archive on a machine other than the smurf server. For example, my timestreams on simons1 are not located at /data/so/timestreams.

dpdutcher:
Yeah my first instinct was to go with default `base_dir=None`, but that kind of clashed with how it was implemented elsewhere, and I touch less files this way.

PR #377, opened 2023-08-07
Adds option to disable compression

jlashner:
option to disable compression to get bias steps working at site.

### DIFF
--- a/sodetlib/operations/bias_steps.py
+++ b/sodetlib/operations/bias_steps.py
@@ -1044,6 +1044,10 @@ def take_bgmap(S, cfg, bgs=None, dc_voltage=0.3, step_voltage=0.01,
             function.
         g3_tag: string, optional
             Tag to attach to g3 stream.
+        enable_compression: bool, optional
+            If True, will tell the smurf-streamer to compress G3Frames. Defaults
+            to False because this dominates frame-processing time for high
+            data-rate streams.
     """
     if bgs is None:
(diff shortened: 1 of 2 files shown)


sodetlib/operations/bias_steps.py
```diff
@@ -1170,7 +1170,7 @@ def take_bias_steps(S, cfg, bgs=None, step_voltage=0.05, step_duration=0.05,
 
         bsa.sid = sdl.stream_g3_on(
             S, tag=g3_tag, channel_mask=channel_mask, downsample_factor=1,
-            filter_disable=True, subtype=stream_subtype
+            filter_disable=True, subtype=stream_subtype, enable_compression=False
```
dpdutcher, line 1173: Should `enable_compression` be a kwarg in take_bias_steps that's passed along to stream_g3_on instead of just always being set to False here?

dpdutcher: does it cause the data on disk to be much bigger such that always running these with compression off could become an issue in labs?

jlashner: Checking on the server, I believe the uncompressed bias step data is ~700 MB, which is large enough that we should be conscious of it in labs... I'll add the option to enable compression but I'm going to leave the default as False.

jlashner: done

dpdutcher (approved):

PR #380, opened 2023-08-14
Add try/except to np.polyfit

dpdutcher:
np.polyfit can fail for some complicated reasons, and it's better to skip a problematic channel than to fail the whole algorithm. I think a try/except here is the easiest solution.

I also don't know all the failure modes (TypeError, SystemError, etc.), so I left it as a generic except.


jlashner (approved):
Thanks daniel, currently looking into why this is failing, but this seems like a necessary change for the time being.

msilvafe:
Should we leave this except in here even after Jack tracked down the cause and merged the fix?

jlashner:
I think it's good to keep this, we really don't want the analysis function to fail out if we can avoid it, since then you lose the iva object. Its much easier to inspect bad channels and failures if the function finishes and returns the iva object

PR #383, opened 2023-08-24
Adds dev cfg option to disable compression always

jlashner:
There was a discussion on the daq-dev call yesterday, and we decided that since we are not retaining L2-data  on site, it is best to disable compression entirely until it is bound into books. This should speed up general data processing time for the streamer and the bookbinder, and should make things like magpie and lyrebird less compute intensive.

This PR just makes `enable_compression` a device cfg option instead of always defaulting to True.


dpdutcher (approved):
Looks fine to me

PR #384, opened 2023-08-30
bump so_smurf_base docker

jlashner:
bump version of so_base_docker. This should fix `pysmurf.__version__` calls, which is being used by specific pysmurf functions to determine behavior.


msilvafe (approved):
Looks good let's get things updated and running out in the world

PR #385, opened 2023-09-01
Adds downsample_factor to packet drop function

jlashner:
Small change to add downsample_factor as a parameter to the packet drop function. Useful for checking that timing is working and packets are cosampled across slots.


msilvafe (approved):
Great to make these QC functions more robust. They’ll be very helpful.

PR #388, opened 2023-09-03
Adds relock_tracking_setup after setup_tracking_params finishes

jlashner:
As @yuhanwyhan and @yaqiongl described in #372, `setup_tracking_params` leaves you in a bad state, where the tracking is only optimized for the last band run, which can lead to high noise on AMC0 since the `lms_freqs` may not be set correctly.

I believe the best way to fix this is to run `relock_tracking_setup` after finding the tracking params for each band, which will chose a `frac_pp` based on the optimal params for each of the running bands, and will adjust `lms_freq` for each band accordingly. I have been running this manually whenever I setup_tracking_params and I think it works well, but it makes sense to just build it into the function.


msilvafe (approved):
I agree with your changes. And given you've been running like this for awhile I think it's safe to merge.

PR #398, opened 2023-12-04
Fixes G3 stream tagging in biasstep_rebias

### LINKED ISSUE #397: Bias step rebiasing tagged incorrectly
Looking at observations starting to come out of satp3, looks like rebiasing with bias steps isn't tagging things the way we'd like. They're all tagged as `oper,bias_steps,oper,biasstep_rebias`. We'll want to fix this because it's going to create a whole lot of unnecessary books

jlashner:
Changes biasstep_rebias so it does proper stream subtyping. Resolves #397


kmharrington (approved):
thanks!

PR #401, opened 2023-12-19
IV param updates

jlashner:
On daq-support, it came up that for some operations such as IV and bias steps, it would be useful to have the ability to have "defaults" which were system or device dependent. 

Max and I came up with this solution for how this may be done for IVs, and can apply it to other fn's if needed. It moves the IV configuration vars in take_iv from just being function params in the local namespace to being held in the `IVCfg` dataclass. This configuration structure has a number of advantages, but most importantly for this use case it allows us to combine cfg params from a number of different sources, explicitly setting the hierarchy. Here, the cfg params will be taken in the following order:
1. be taken from `**kwargs` passed in through `take_iv`
2. use kwargs in `cfg.dev.exp['iv_defaults']`
3. Use default kwargs of the IVCfg class

This will need to be tested before merging

### DIFF
--- a/sodetlib/det_config.py
+++ b/sodetlib/det_config.py
@@ -42,6 +42,9 @@ exp_defaults = {
     'downsample_mode': 'internal',
     'enable_compression': True,
 
+    # Dict of device-specific default params to use for take_iv
+    'iv_defaults': {},
+
     # Amp stuff
     "amps_to_bias": ['hemt', 'hemt1', 'hemt2', '50k', '50k1', '50k2'],
     "amp_enable_wait_time": 10.0, "amp_step_wait_time": 0.2,
--- a/sodetlib/operations/iv.py
+++ b/sodetlib/operations/iv.py
@@ -5,6 +5,8 @@ import sodetlib as sdl
 import matplotlib.pyplot as plt
 import os
 from tqdm.auto import trange
+from typing import Optional
+from dataclasses import dataclass, asdict
 
 import warnings
 warnings.filterwarnings('ignore')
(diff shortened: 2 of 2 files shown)


ykyohei:
@msilvafe and @ykyohei did take_iv in the following two conditions and confirmed that the IV curves are taken using identical paeameters

```
iv_study_all= iv.take_iv(
    S, cfg, bias_groups=list(range(12)), wait_time=0.01, bias_high=18,
    overbias_wait=2, bias_low=0, bias_step=0.025, overbias_voltage=19,
    cool_wait=20, high_current_mode=False, run_serially=True,
    serial_wait_time=30, run_analysis=True, show_plots=True,
)
cfg.dev.exp['iv_defaults'] = {}

/so/level2-daq/satp3/smurf/17030/ufm_mv5/1703082728_take_iv/    

iv_study_all= iv.take_iv(S, cfg)
cfg.dev.exp['iv_defaults'] = {'bias_groups': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
 'wait_time': 0.01,
 'bias_high': 18,
 'overbias_wait': 2,
 'bias_low': 0,
 'bias_step': 0.025,
 'overbias_voltage': 19,
 'cool_wait': 20,
 'high_current_mode': False,
 'run_serially': True,
 'serial_wait_time': 30,
 'run_analysis': True,
 'show_plots': True}

/so/level2-daq/satp3/smurf/17030/ufm_mv5/1703084556_take_iv
```

ykyohei (approved):

jlashner:
Great, thanks kyohei!!

msilvafe:
@ykyohei and I tested this branch. In a notebook on the smurf server we performed the following tests:
1. Confirm that we can run the IV passing all of the args to the fn directly as was done previously
```
iv_study_all= iv.take_iv(
    S, cfg, bias_groups=list(range(12)), wait_time=0.01, bias_high=18,
    overbias_wait=2, bias_low=0, bias_step=0.025, overbias_voltage=19,
    cool_wait=20, high_current_mode=False, run_serially=True,
    serial_wait_time=30, run_analysis=True, show_plots=True)
```
2. Update the device config `iv_defaults` key to contain the above arguments and then run as:
```
iv_study_all= iv.take_iv(S, cfg)
```
3. Inspect the plots and iv_analysis files to confirm that these two methods are producing the same results (see below):

| Args from device config  | Passing All Args |
| ------------- | ------------- |
| image (15) | image (17) |
| image (14) | image (16) |

And here's comparing the parameters in the IV analysis files:
```
bias_groups [ 0  1  2  3  4  5  6  7  8  9 10 11] [ 0  1  2  3  4  5  6  7  8  9 10 11]
overbias_voltage 19 19
overbias_wait 2 2
high_current_mode False False
cool_wait 20 20
cool_voltage None None
min biases 1.0231815394945443e-12
max biases 18.0 18.0
bias_high 18 18
bias_low 0 0
bias_step 0.025 0.025
show_plots True True
wait_time 0.01 0.01
run_analysis True True
run_serially True True
serial_wait_time 30 30
g3_tag None None
analysis_kwargs {} {}
```

So I think this is probably good to merge!

msilvafe (approved):
Jack + I co reviewed this as he was writing it on a slack huddle and then we tested it with Kyohei on satp3 so I'm happy to merge.

jlashner:
Perfect, thanks for the extensive testing

PR #404, opened 2024-01-03
Adds dev-cfg defaults for bias_steps and take_bgmap

jlashner:
This does the same thing we did for the take_iv function, moving configuration params for `take_bias_steps` and `take_bgmap` into separate dataclasses so they can be more easily updated with defaults from the device cfg. This also simplifies a bit the storage of run config params and the passing of shared params between take_bgmap and take_bais_steps.

This has not yet been tested.


msilvafe (approved):
This looks good, once we get to test this I'm happy to merge.

dpdutcher:
Done in #479

PR #407, opened 2024-01-09
Better determination of det transition state in bias steps

jlashner:
This branch implements Yaqiong's bias-step transition determination code described here: https://github.com/simonsobs/sodetlib/discussions/361#discussioncomment-6442939

It's been tested on the LAT and fixes biasing isuses, so safe to merge.


msilvafe (approved):
Happy we're getting this in and thanks for testing on lat. Merge away.

PR #411, opened 2024-01-26
Updates links for most recent PCIe fw

jlashner:
Updates links and docs in the `load_pcie_fw` folder so that it loads most recent binaries.


msilvafe:
Where are the changes to the firmware discussed/documented? I think maybe we should raise this to @dpdutcher to disseminate the updates to all the systems?

PR #418, opened 2024-02-23
Remove rsync docs deployment workflow

BrianJKoopman:
This has been broken ever since SSH restrictions put in place. Just making way for setup on readthedocs.


BrianJKoopman:
Just pinging again about this. Actions deploying the HTML build continue to be broken.

msilvafe (approved):

Issue #416, opened 2024-02-07
Fix responsivity in iv


satoru99:
There are two issues.
**1. Calculation of R_L**
The load resistance (R_L) should be shunt resistance (R_sh) + parasitic resistance (R_par) as written in comment.
It is calculated as follows
https://github.com/simonsobs/sodetlib/blob/1b384e513d3d4a3d4bfd8282f378b29e065c591b/sodetlib/operations/iv.py#L395-L397
However, from the equation of voltage:
`R_sh * I_bias = (R_tes + R_sh + R_par) * I_tes`.
the `R` evaluated here should be R_tes + R_par,
and, the `R_L` only contains R_par and not R_sh.
(Also, `R_n` might include contribution from R_par.)
To give the expected R_L, we should add `+ R_sh`.

**2. Calculation of dV/dI**
The responsivity is calculated as follows. Note that the `rL` comes from `R_L` above and thus underestimated by R_sh.
https://github.com/simonsobs/sodetlib/blob/1b384e513d3d4a3d4bfd8282f378b29e065c591b/sodetlib/operations/iv.py#L299
Here, dv is calculated from `v_tes`.
But from the theoretical model, the correct value seems to be the derivative of `v_bias`.
Since, `v_tes = v_bias * R / (R + R_L)`, this difference affects at small R. 

These two issues affect the responsivity calculation. It's necessary to be fixed.

msilvafe:
Thanks @satoru99, yes we (I) seem to have dropped `R_par` completely from all of these equations and we did not use the thevenin equivalent voltage (`i_bias*R_sh`) as you pointed out here. I follow your math/logic (from Irwin-Hilton) and agree that we should fix this. These seem like the are pretty simple changes to make, and I would love to see things tightening up at small R from this. Are you planning to show some example re-analyzed IVs and submit a PR with these fixes or work with @RemingtonGerras to do this?

RemingtonGerras:
[Confluence page of notes + summary of the PR](https://simonsobs.atlassian.net/wiki/spaces/~638a5b0b8fd2d2d5f13165d5/pages/374505518/Detector+Calibration+using+RP+Curve)

PR #424, opened 2024-04-04
Addresses issue #416 AND recalculates bias step TES parameter values 

RemingtonGerras:
Bias Step TES parameter values are calculated from the RP curve that satoru worked on!

### DIFF
--- a/sodetlib/operations/translate_tes_parameters.py
+++ b/sodetlib/operations/translate_tes_parameters.py
@@ -1,0 +1,569 @@
+import numpy as np
+from scipy.interpolate import interp1d
+from scipy.optimize import curve_fit
+from sodetlib.operations.iv import IVAnalysis
+from sodetlib.operations.bias_steps import BiasStepAnalysis
+from tqdm.auto import trange
+
+
+import warnings
+warnings.filterwarnings('ignore')
+np.seterr(all="ignore")
+
+
+def translate_tes_params(iv_file, bs_file, save_iv=False, save_bs=False):
+    """
+    Driver function that recalculates IV TES parameters using the
+    thevenin equivalent voltage. Then computes the RP curve following
+    Satoru's methods. Finally, load in and recalculate bias step analysis
+    TES parameters using a shifted RP curve determined by the change in 
+    optical power between operations. 
+    
+    Args:
+    -----
+    iv_file : str
+        filepath to iv_analysis.npy file to re-analyze.
+        
+    bs_file : str
+        filepath to bs_analysis.npy file to re-analyze
+    
+    save_iv : bool
+        Boolean that determines if the re-analyzed iva is saved
+        
+    save_bs : bool
+        Boolean that determines if the re-analyzed bsa is saved
+    """
+    iva2 = reanalyze_iv(iv_file, save_iv)
+    bsa2 = rerun_analysis(bs_file, iva2, save_bs)
+    
+    return iva2, bsa2
+
+
+###############################################################################
+# IVA translation functions below:
+# 1. Calculate TES parameters using thevenin equivalent voltage.
+# 2. Compute RP curve and storing parameters that determine RP curve.
+# 3. Save updated IVAnalysis object as a numpy file.
+###############################################################################
+
+def model_logalpha(r, logp0=0., p1=7., p2=1., logp3=0.):
+    """
+    Fits Alpha' parameter. P0 has units of [uW]^-1.
+    """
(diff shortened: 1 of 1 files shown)


sodetlib/operations/translate_tes_parameters.py
jlashner, line 1: This needs either a link to Satoru's confluence page describing the math, or better yet an RST page in the sodetlib docs dedicated to this analysis.

sodetlib/operations/translate_tes_parameters.py
```diff
@@ -0,0 +1,569 @@
...
+        P0 is optical power
+    """
+    nchans = iva['nchans']
+    popt_size = 6
+    popts = np.full((nchans, popt_size), np.nan)
+    
+    r1 = np.linspace(min(minrfrac, 0.0001), max(maxrfrac, 0.9999), 9999)
+    dr = r1[1] - r1[0]
+
+    def model_Pj(r, logp0=0., p1=7., p2=1., logp3=1.,
```
jlashner, line 79: Can you document here the mathematical description of the model function (or reference satoru's page), and also document the params?

sodetlib/operations/translate_tes_parameters.py
```diff
@@ -0,0 +1,569 @@
...
+from sodetlib.operations.bias_steps import BiasStepAnalysis
+from tqdm.auto import trange
+
+
+import warnings
+warnings.filterwarnings('ignore')
+np.seterr(all="ignore")
+
+
+def translate_tes_params(iv_file, bs_file, save_iv=False, save_bs=False):
```
jlashner, line 14: Instead of passing file paths, can you accept pre-loaded iva and bsa dicts? This is more versatile and plugs in better with how we use it in the det_cal scripts.

jlashner, line 14: Also you don't need to worry about saving files, so I'd remove the args and all logic pertaining to saving the data.

jlashner: Returned values (iva2, bsa2) should start as deepcopies of the provided dicts.

sodetlib/operations/translate_tes_parameters.py
```diff
@@ -0,0 +1,569 @@
...
+    iva2['p_sat_V1'] = np.full(iva2['p_sat'].shape, np.nan)
+    
+    for i in range(iva2['nchans']):
+        if np.isnan(iva2['R_n_V1'][i]):
+            continue
+
+        level = psat_level
+        R = iva2['R_V1'][i]
+        R_n = iva2['R_n_V1'][i]
+        p_tes = iva2['p_tes_V1'][i]
```
jlashner, line 139: For these, and all other corrected values in both the iva and bsa objects, can you replace the existing keys (of the new deepcopied dicts) instead of adding new ones?

sodetlib/operations/translate_tes_parameters.py
```diff
@@ -0,0 +1,569 @@
...
+        np.save(fname, iva2, allow_pickle=True)
+        
+###############################################################################
+# BSA translation functions below:
+# 1. Fit for delta_Popt by fitting the associated IVA RP_curve to 
+#     the measured dI/dIb from BSA.
+# 2. Recompute R, Si, Pj, and I0 based on RP_cuve shifted by delta_Popt
+# 3. Save updated BSAnalysis object as a numpy file.
+###############################################################################
+def rerun_analysis(bs_file, iva2=None, save=False, R0_thresh=30e-3):
```
jlashner, line 325: iva2 should just be a positional argument

sodetlib/operations/translate_tes_parameters.py
```diff
@@ -0,0 +1,569 @@
...
+        filepath to bs_analysis.npy file to re-analyze
+    
+    save_iv : bool
+        Boolean that determines if the re-analyzed iva is saved
+        
+    save_bs : bool
+        Boolean that determines if the re-analyzed bsa is saved
+    """
+    iva2 = reanalyze_iv(iv_file, save_iv)
+    bsa2 = rerun_analysis(bs_file, iva2, save_bs)
```
jlashner, line 37: To both the iva and bsa objects, can you add a new key `version` which should be set to 1? Then this can be checked to make sure corrections aren't re-applied to already corrected results.

jlashner (changes requested):
Hi Remy, thanks for this!! Its very nice code, and works well for the testing I've done on existing observations. I have some inline comments with things that are coming up as I'm testing with existing data, and thinking about how it should integrate with the det_cal db script.

PR #425, opened 2024-04-30
Change stream_g3_on kwarg make_freq_mask default to False

jlashner:
The `make_freq_mask` function involves a lot of EPICS calls and takes  a long time to run. This is currently causing some [issues on SAT1](https://github.com/simonsobs/daq-discussions/discussions/80), and as far as I know the resulting "freq.txt" file is not use anywhere in SO. 

This changes the stream defaults so that it does not attempt to make a frequency mask.

### DIFF
--- a/sodetlib/stream.py
+++ b/sodetlib/stream.py
@@ -127,7 +127,7 @@ def take_g3_data(S, dur, **stream_kw):
 
 
 @set_action()
-def stream_g3_on(S, make_freq_mask=True, emulator=False, tag=None,
+def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
                  channel_mask=None, filter_wait_time=2, make_datfile=False,
                  downsample_factor=None, downsample_mode=None,
                  filter_disable=False, stream_type=None, subtype='stream',


msilvafe (approved):
Good investigation, are you sure that none of the data packaging operations use the mask file that's written? Is there a test we can run off this branch before we merge to check double check that nothing downstream needs that? If you've checked and are pretty confident that nothing downstream uses this I'm ok just merging.

jlashner:
Just checked and this is the only instance of `freq.txt` in all of sotodlib, so I think we should be ok to remove it without testing: https://github.com/simonsobs/sotodlib/blob/1e85b2c3a41c329cb396cd278a81fa3a05e48ee7/sotodlib/io/imprinter.py#L677-L679

msilvafe:
> Just checked and this is the only instance of `freq.txt` in all of sotodlib, so I think we should be ok to remove it without testing: https://github.com/simonsobs/sotodlib/blob/1e85b2c3a41c329cb396cd278a81fa3a05e48ee7/sotodlib/io/imprinter.py#L677-L679

Ok thanks for double checking.

PR #426, opened 2024-05-13
move bias_step init_current_mode outside of try statement

jlashner:
This moves the `init_current_mode` function outside of the try statement. This is a function to get the initial state of the current-mode, so we can recover. If this statement fails, since we haven't yet changed state, we don't need to run the `finally` block that recovers the initial state, so we don't need it inside of the `try` block.

Currently there are some instances where this function fails, and then the state-recovery fails since `init_current_mode` is not defined, such as in [this discussion](https://github.com/simonsobs/daq-discussions/discussions/92).


msilvafe (approved):
Thanks for the quick fix.

PR #427, opened 2024-06-05
Update bias_dets.py

dpdutcher:
In biasstep_rebias: it should be `cfg.dev.exp`, not `cfg.dev.experiment`

### DIFF
--- a/sodetlib/operations/bias_dets.py
+++ b/sodetlib/operations/bias_dets.py
@@ -307,7 +307,7 @@ def biasstep_rebias(
         
         sdl.overbias_dets(S, cfg, bias_groups=bg_overbias_needed)
         ##force wait for UFM to reach equalibrium 
-        sleep_time = cfg.dev.experiment["overbias_sleep_time_sec"]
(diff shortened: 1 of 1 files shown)


msilvafe:
I'm a bit confused why you needed this change. The device config code loads everything in the `experiment` section of the dev_cfg yaml into `cfg.exp`: https://github.com/simonsobs/sodetlib/blob/514b8b16042ab10496639b89d4fb98154cfa7495/sodetlib/det_config.py#L114-L172

dpdutcher:
cfg.dev does not have an "experiment" attribute, it only has a "exp" attribute. 
So `sleep_time = cfg.dev.experiment["overbias_sleep_time_sec"]` fails with an attribute error.

msilvafe:
> cfg.dev does not have an "experiment" attribute, it only has a "exp" attribute. 
> 
> So `sleep_time = cfg.dev.experiment["overbias_sleep_time_sec"]` fails with an attribute error.

Ahh ok I read this wrong I thought you were changing exp -> experiment not the other way around.

msilvafe (approved):
Thanks for testing and catching this.

Issue #393, opened 2023-10-20
Psat not calculated in most robust way from IV curves


dpdutcher:
As I understand it, IV data is flipped during analysis, so that the first element in an array will be the last datapoint that was taken, and vice-versa (see this comment) https://github.com/simonsobs/sodetlib/blob/c7bc19793ccfd168435885930d819fb58a506e8e/sodetlib/operations/iv.py#L326
 In one sense that makes looking at the data confusing, because it does not match the actual time order, but in another sense more intuitive, because the left-to-right direction on most IV plots matches the start-to-finish order of the data arrays, so, fine.
Where it does run into a problem is computing Psat values. Currently, the code looks for the first time R/R_n is above psat_level:
https://github.com/simonsobs/sodetlib/blob/c7bc19793ccfd168435885930d819fb58a506e8e/sodetlib/operations/iv.py#L218-L228

But because IV curves can be ill-behaved in the low voltage, superconducting portion of the IV curve, which in this analysis scheme comes first, the code should instead look for _the last time R/R_n is below psat_level_. So something like

```
        cross_idx = np.where(R/R_n < level)[0]

        if len(cross_idx) == 0:
            iva.p_sat[i] = np.nan
            continue

        # Takes cross-index to be the last time Rfrac crosses psat_level
        cross_idx = cross_idx[-1]
        if cross_idx == 0:
            iva.p_sat[i] = np.nan
            continue
```

I have yet to make and run with this change to make sure this fixes the problem, but thought I'd raise this issue.

dpdutcher:
I finally got around to addressing this issue. The plot below illustrates the problem, where the p_sat calculation is thrown-off by an instability in the low-R_tes portion of the IV curve. Note that target bias voltages are not affected by this.

![image](https://github.com/user-attachments/assets/44c1bb44-6563-4629-aef8-3d8bab7ff819)
Also, the full P_sat distribution looks like the below. Note all the values between 0 pW and 16 pW
![image](https://github.com/user-attachments/assets/78ea5767-0faf-434a-b5ff-b8514a1aa60a)

I then made the committed changes to the code, switching the logic from the first time R is above the target r_frac to the last time R is below the target r_frac. This fixed the psat determination in the selected detector, and it also fixed the total psat histogram to get rid of those spurious low values.

![image](https://github.com/user-attachments/assets/9e96873e-7cc1-4f90-9858-accecdd00434)
[...]

PR #435, opened 2024-08-09
Change compute_psats to use last crossing of Rfrac=target_level

dpdutcher:
I finally got around to addressing #393 . The plot below illustrates the problem, where the p_sat calculation is thrown-off by an instability in the low-R_tes portion of the IV curve. Note that target bias voltages are not affected by this.

image
Also, the full P_sat distribution looks like the below. Note all the values between 0 pW and 16 pW
image

I then made the committed changes to the code, switching the logic from the first time R is above the target r_frac to the last time R is below the target r_frac. This fixed the psat determination in the selected detector, and it also fixed the total psat histogram to get rid of those spurious low values.

image
image

Tested on a full UHF UFM in a Princeton test cryosat.


msilvafe (approved):
Thanks for the update.

PR #437, opened 2024-08-13
Adds sodetlib version to metadata

### LINKED ISSUE #436: Add sodetlib version to get_metadata
When were changing analysis methods in different on-site operations, we need to know which versions of analysis are in which data files. At a minimum I think we need the sodetlib version in https://github.com/simonsobs/sodetlib/blob/master/sodetlib/util.py#L187 

We should probably also make analysis changes be a tagged release?

jlashner:
This PR adds the sodetlib version tag to the common metadata dict.

Resolves #436.

### DIFF
--- a/sodetlib/util.py
+++ b/sodetlib/util.py
@@ -7,6 +7,7 @@ from scipy import signal
 import time
 import os
 from collections import namedtuple
+import sodetlib
 from sodetlib import det_config
 from sodetlib.constants import *
 from sotodlib.tod_ops.fft_ops import calc_psd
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
is it possible to run this through a docker system to make sure the version that get's added is useful? I've seen dockers do weird things

jlashner:
I did check this in a jackhammer util and it seems to be what we want:
```
cryo@smurf-srv20-satp1:~$ jackhammer util
cryo@smurf-srv20-satp1:/usr/local/src/pysmurf$ python3
Python 3.8.10 (default, May 26 2023, 14:05:08)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sodetlib
sodetlib.__version__
>>> sodetlib.__version__
'0.5.2+21.geaa66cc'
```
But ya I know there were docker/versioning issues with pysmurf previously

kmharrington:
great!

PR #440, opened 2024-09-09
Add fixed tones support

tristpinsm:
This is a draft with a proposed implementation for supporting fixed tones. If this looks good, I can proceed to testing on the SLAC system and then polish things up / add docstrings.

Right now it relies on a new function in `pysmurf` (https://github.com/slaclab/pysmurf/pull/796) because I thought that would be a more complete implementation, but if it's preferable to keep these changes within `sodetlib` I can move it into this PR.


jlashner:
Hi Tristan, this looks like a good start! A few comments:
 - I think (band, channel) are not enough to determine fixed tone location, and we should also include the frequency or the freq relative to subband center. I kind of think just specifying and saving the (frequency, band) or (frequency, AMC) information is easiest because from there you can get the band and channel info.
- I think the "setup_fixed_tones" function should also figure out the optimal frequency of the fixed tones, since we don't expect users to know this. I imagine it should take some behavioral parameters like `fixed_tones_per_band` and `min_fixed_tone_spacing` and then just figure out the frequencies based on that, and save them to the device cfg.
- The defaults for the device cfg are recorded [here](https://github.com/simonsobs/sodetlib/blob/438d2b9d3f0e508b8a65c5d4e59a6170fc0b5054/sodetlib/det_config.py#L37). Anything we add to the device cfg should be added here, and then in functions we can assume these keys will always be present in the cfg. We can make it so the `fixed_tones` entry is an Optional[dict], and only set to a value after the setup function is run.
 
One more thing is that recently I've found type hinting and static type checking to be very helpful for code readability and catching bugs early. For instance, it would be a lot clearer if the `freq` arg of setup_fixed_tones was hinted as `List[float]`, so its clearer that it should be a list and you can't pass a single value. I'm not going to require this for this PR since the rest of sodetlib doesn't follow it, but thought I'd point it out because I think its really nice!

tristpinsm:
Thanks for the rapid feedback! That all makes sense and I'll make those changes

sodetlib/operations/uxm_setup.py
```diff
@@ -363,6 +363,35 @@ def setup_tune(S, cfg, bands, show_plots=False, update_cfg=True):
     return True, summary
 
 
+def setup_fixed_tones(S, cfg, freq, tone_power=10, update_cfg=True):
```
msilvafe, line 366: We probably also need an automated procedure to find the frequencies to use based on gaps where resonances don't exist from the tunefiles or find_freq results.

sodetlib/stream.py
```diff
@@ -207,6 +208,11 @@ def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
     S.set_downsample_factor(downsample_factor)
     S.set_filter_disable(int(filter_disable))
 
+    # ensure fixed tones are on
+    for band in S.bands:
+        if "fixed_tones" in cfg.dev.bands[band]:
+            uxm_setup.turn_on_fixed_tones(S, cfg, band)
```
msilvafe, line 214: I'm not sure this will work since you need to run tracking_setup after turning on fixed tones.

jlashner: I think we should just figure out how to enable fixed tones without tracking setup... I imagine there's just a register or two that needs to be set.

tristpinsm: I had looked through `tracking_setup` and it wasn't clear to me what would be missing, but in any case I will have to go through all of these steps with testing and I should find out then.

tristpinsm:
I just pushed a new draft addressing some of the suggestions so far:
- Now it sets the channel frequency offsets
- These are saved along with the channels in the config
- I cobbled together an algorithm for picking fixed tone frequencies, but I guess we should consider it a placeholder until I can do some actual testing

After realising that you need the subband frequency offsets, I moved all the steps into the `sodetlib` functions, as I didn't see an obvious series of steps to do this within the existing `pysmurf` functions.

jlashner:
Sorry should've mentioned this, there is already an algorithm in pysmurf that we tend to use when doing this manually:
https://github.com/slaclab/pysmurf/blob/1ff66325bf174f975c241b5cf6922f378dd58b20/python/pysmurf/client/util/smurf_util.py#L4361

jlashner:
We may want to make something custom though that is based on an existing tunefile and doesn't need to rescan.

tristpinsm:
> Sorry should've mentioned this, there is already an algorithm in pysmurf that we tend to use when doing this manually: https://github.com/slaclab/pysmurf/blob/1ff66325bf174f975c241b5cf6922f378dd58b20/python/pysmurf/client/util/smurf_util.py#L4361

Oh man, I hadn't seen that! I'll take a look and see how it could fit in.

sodetlib/stream.py
tristpinsm: I added a `turn_on_fixed_tones` call to `relock_tracking_setup` so that if it disables these channels they should be restored

tristpinsm:
I tried to address the suggestions so far and did some testing on the SLAC UFM today. I think functionally things are working now, and next I will polish up the code and commit history and add docstrings.

sodetlib/operations/uxm_setup.py
```diff
@@ -363,6 +363,162 @@ def setup_tune(S, cfg, bands, show_plots=False, update_cfg=True):
...
+    update_cfg : bool
+        Whether to save the fixed tone configuration to the device config. Defaults to True.
+
+    Returns
+    -------
+    fixed_tones : dict
+        Fixed tones configuration, as saved to device config.
+    """
+
+    bands = cfg.dev.exp['active_bands']
```
jlashner, line 425: Can you accept a `bands` arg of type Optional[List] which defaults to using active_bands?

sodetlib/operations/uxm_setup.py
```diff
@@ -363,6 +363,162 @@ def setup_tune(S, cfg, bands, show_plots=False, update_cfg=True):
...
+        for band, data in fixed_tones.items():
+            cfg.dev.update_band(band, {"fixed_tones": data})
+
+    S.log("Done setting fixed tones.")
+
+    return fixed_tones
+
+
+@sdl.set_action()
+def turn_on_fixed_tones(S, cfg, band):
```
jlashner, line 476: Can this also default to running on a group of bands (defaulting to active_bands) instead of just one? That is how we will always be running in operation so its easier if we don't need to loop over bands every time.

sodetlib/operations/uxm_setup.py
```diff
@@ -363,6 +363,162 @@ def setup_tune(S, cfg, bands, show_plots=False, update_cfg=True):
...
+        ntone = num_gaps
+
+    # return the central frequencies in selected gaps
+    gap_ind = np.argsort(gaps)[:-(ntone + 1):-1]  # ntone largest gaps
+    return 0.5 * (res_freq[gap_ind + 1] + res_freq[gap_ind])
+
+
+@sdl.set_action()
+def setup_fixed_tones(
+    S, cfg, fixed_tones_per_band=8, min_gap=10, tone_power=10, update_cfg=True
```
jlashner, line 397: Is there any reason the default tone_power is 10 instead of the default in the band config (we typically use 12)? I don't see any reason the fixed-tone power should be the same as the resonator tone-power, just wondering if there was any reason behind the default.

sodetlib/stream.py
jlashner: I think we still need an additional `enable_fixed_tones` cfg entry which allows us to disable fixed tones without removing the fixed-tone dictionary.

jlashner (changes requested):
Thanks for the changes tristan! I have a few more simple requests and then I'm happy to merge (assuming its been tested)!

tristpinsm:
> Thanks for the changes tristan! I have a few more simple requests and then I'm happy to merge (assuming its been tested)!

Thanks for the suggestions -- I'll make those changes. I've tested the individual functions, but not in a full OCS context with agents running. But I did want to try to get something like that working.

sodetlib/operations/uxm_setup.py
msilvafe: The fixed tones don't get attenuated by the resonator dip before hitting the amplifier so they contribute much more to the total nonlinearity than the locked tones. The adjustment down by 6dB is supposed to set them at roughly equal power as the locked channels.

sodetlib/operations/uxm_setup.py
```diff
@@ -363,6 +363,162 @@ def setup_tune(S, cfg, bands, show_plots=False, update_cfg=True):
     return True, summary
 
 
+def _find_fixed_tone_freq(S, band, ntone, min_spacing, band_start=-250, band_end=250):
+    """Place fixed tones in the largest gaps between resonators."""
```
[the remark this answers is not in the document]
msilvafe: Types is not a requirement right?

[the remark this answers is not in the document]
jlashner: It's not, but it would be helpful for me and others reading if it were at least specified in the docstring if not in the signature.

tpsatt:
Thanks for all this work @tristpinsm! Wondering if it's worth adding a call to `turn_on_fixed_tones` in `uxm_relock` that way this change can fit nicely into our current schedules?

Also should feedback be disabled in `tracking.setup_tracking_paramters` too in case `relock_tracking_params` doesn't get called?

tristpinsm:
> Thanks for all this work @tristpinsm! Wondering if it's worth adding a call to `turn_on_fixed_tones` in `uxm_relock` that way this change can fit nicely into our current schedules?
The way it is now, it is called in `stream_g3_on` in order to guarantee the fixed tones are set during data-taking. Are you thinking of scenarios where fixed tones would need to be set but we are not streaming data? Like an interactive session or something?

> Also should feedback be disabled in `tracking.setup_tracking_paramters` too in case `relock_tracking_params` doesn't get called?
`relock_tracking_params` is called as the last step of `setup_tracking_params` so I think this is covered.

It doesn't look like I got around to changing the code for finding fixed tone channels to use existing functions -- I can take another look at that.

msilvafe:
> Ah I'd missed both of these points. Thanks @tristpinsm! Looks good on my end to rebase and merge

Let me take one last look over if that's ok @tpsatt and @tristpinsm.

sodetlib/stream.py
```diff
@@ -207,6 +208,9 @@ def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
     S.set_downsample_factor(downsample_factor)
     S.set_filter_disable(int(filter_disable))
 
+    # ensure fixed tones are on
+    uxm_setup.turn_on_fixed_tones(S, cfg)
```
msilvafe, line 212: Can we check `cfg.dev.bands[b]['fixed_tones']['enabled']` and only run this if True?

msilvafe (approved):
This looks good, I think we should fix the broken sphinx docs build also but probably not a requirement here.

sodetlib/stream.py
tristpinsm: it checks this band-by-band within `turn_on_fixed_tones`. We could check it here as well, but the result would be the same.

sodetlib/operations/tracking.py
```diff
@@ -449,6 +449,53 @@ def setup_tracking_params(S, cfg, bands, update_cfg=True, show_plots=False):
...
+    ft_ind = {}
+    for band in bands:
+        amp = S.get_amplitude_scale_array(band)
+        feedback = S.get_feedback_enable_array(band)
+        ft_ind[band] = np.where((amp != 0) & (feedback == 0))[0]
+
+    return ft_ind
+
+
+def _restore_fixed_tone_channels(S, chan_per_band: dict):
```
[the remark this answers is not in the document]
tristpinsm: I think the idea here was to restore the state that was present before running `relock_tracking_setup`, whatever that state was -- not necessarily the state that is stored in the config. This is to avoid what could be a confusing outcome if you relock the tracking and suddenly the fixed tones configuration has changed.

But if this doesn't sound right, we should change it.

tristpinsm:
thanks for approving and the suggestion @jlashner !

I've implemented that now, so when we're ready to test @msilvafe we should make sure to pull this commit.

PR #442, opened 2024-09-24
Clear data script

jlashner:
This PR implements a clear-data script to scan directories and delete data from smurf server according to the [data deletion proposal](https://simonsobs.atlassian.net/wiki/spaces/~55705801070cc775254537ab15f2dcf6702b78/pages/360448021/2024-03-28+Smurf+Data+Deletion+Proposal).

It will scan directories from which we want to delete data, and print directories to be deleted before asking the user if it should continue. For testing, it can be run in "dry" mode to see which removal commands will be run, like:
```
python3 clear_old_data.py --dry
```
To actually remove the files, run without the `--dry` argument.

This has been tested on smurf-so3 and smurf-srv20.

### DIFF
--- a/scripts/clear_old_data.py
+++ b/scripts/clear_old_data.py
@@ -15,12 +15,24 @@ python3 clear_old_data.py --dry
 from dataclasses import dataclass
 import os
 import datetime
-from typing import List, Optional
+from typing import List, Optional, Dict, Any
 from enum import Enum, auto
 import yaml
 import shutil
 import argparse
+import logging
+from copy import deepcopy
 
+logger = logging.getLogger()
+
+_nameToLogLevel = {
+    'DEBUG': logging.DEBUG,
+    'INFO': logging.INFO,
+    'WARN': logging.WARNING,
+    'WARNING': logging.WARNING,
+    'ERROR': logging.ERROR,
+    'CRITICAL': logging.CRITICAL,
+}
 
 @dataclass
 class Config:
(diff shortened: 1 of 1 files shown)


scripts/clear_old_data.py
```diff
@@ -0,0 +1,179 @@
...
+@dataclass
+class Config:
+    dry: bool = True
+    delete_smurf_data_after_days: int = 31
+    delete_timestream_data_after_days: int = 365 // 2
+    delete_core_dumps_after_days: int = 365
+    delete_logs_after_days: int = 365
+    verbose: bool = True
+
+    @classmethod
```
msilvafe, line 20: Please give an example cfg file and usage from cli w/ argparse in some docs.

scripts/clear_old_data.py
```diff
@@ -0,0 +1,179 @@
...
+def scan_log_dirs(cfg: Config) -> List[FileInfo]:
+    log_dir = '/data/logs'
+    files: List[FileInfo] = []
+    now = datetime.datetime.now()
+    max_time_delta = datetime.timedelta(
+        days=cfg.delete_logs_after_days
+    )
+    for f in os.listdir(log_dir):
+        path = os.path.join(log_dir, f)
+        ts = os.path.getmtime(path)
```
msilvafe, line 128: Why are you using `getmtime` here and `getctime` in the core dumps?

scripts/clear_old_data.py
```diff
@@ -0,0 +1,179 @@
...
+    files_to_delete: List[FileInfo] = []
+    files_to_delete += scan_smurf_data(cfg)
+    files_to_delete += scan_timestream_dirs(cfg)
+    files_to_delete += scan_core_dumps(cfg)
+    files_to_delete += scan_log_dirs(cfg)
+
+    print(f"{len(files_to_delete)} files to delete:")
+    for f in files_to_delete:
+        days_old = (now - f.dt).days
+        print(f' - {f.path} ({days_old} days old)')
```
msilvafe, line 161: Can we log these to watch in Loki?

msilvafe (approved):
Thanks Jack, very basic doc or one main docstring explaining yaml vs cli usage and a few inline comments/questions but I've approved.

scripts/clear_old_data.py
jlashner: There is not a huge difference, but I chose the mod-time here because the log-path is a directory, and I wanted the timestamp to be the time at which the last file was created or modified, instead of the creation time of the directory.

msilvafe: Ahh gotcha, makes sense.

scripts/clear_old_data.py
jlashner: I don't really want to add loki support... it adds a requirement of network connectivity and making sure the hostnames are set correctly that I think would be a pain to deal with across servers. How about if we allow logging to a file in addition to stdout?

msilvafe: Yup to file sounds good! Would just like to find a record somewhere of what we deleted when.

PR #444, opened 2024-10-01
Adds data as a volume mount to run_gui

jlashner:
Mounts in the `/data` directory to the run_gui script. This was required for shawn to save state from the gui.


msilvafe (approved):
Thanks!

Issue #445, opened 2024-10-04
resonator_fitting: frequency spacing plotted wrong


dpdutcher:
In the seldom-used resonator_fitting.py, the function `plot_fit_summary` incorrectly plots the frequency spacing of the resonances. This is because (1) the frequencies returned from `fit_tune` are physical frequencies (4--6 GHz)  (2) results from all bands are returned as one array (3) `sorted()` is run on the entire array before `np.diff()` is run, mixing resonances that are on separate RF transmission lines.

PR #446, opened 2024-10-04
Fix resonator frequency spacing plot

dpdutcher:
Addresses #445 

The old code ran np.quantile on the list of frequencies, then sorted it, then took the diff.
The new code takes the diff of the list of frequences, which is already sorted in frequency band order, and runs np.quantile on that.

Note this isn't perfect, as it still treats the two separate RF transmission lines (bands 0--3 and bands 4--7) as one, but the one -2GHz spacing is cut out via use of np.quantile.

Tested in Princeton R+D testbed.

Before:
image

After:
image

### DIFF
--- a/sodetlib/resonator_fitting.py
+++ b/sodetlib/resonator_fitting.py
@@ -239,7 +239,7 @@ def fit_tune(tunefile):
     scan_freqs = []
     res_freqs = []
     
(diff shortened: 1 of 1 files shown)


msilvafe (approved):
Thanks for finding + fixing this bug. Please close the issue when you merge.

PR #447, opened 2024-10-17
Adds support for custom pysmurf-controller docker service names

jlashner:
This PR adds support for setting docker service names for pysmurf-controller dockers. This will enable us to change docker services from `ocs-pysmurf-s<slot>` to `ocs-<instance-id>` by specifying manually the service names to use in the sys-config.

Related: https://github.com/simonsobs/ocs-deployment-configs/issues/126#issuecomment-2418366911

### DIFF
--- a/hammers/jackhammer
+++ b/hammers/jackhammer
@@ -36,6 +36,16 @@ sys_config_file = os.path.join(os.environ['OCS_CONFIG_DIR'], 'sys_config.yml')
 with open(sys_config_file, 'r') as stream:
     sys_config = yaml.safe_load(stream)
 
+
+def get_pysmurf_controller_docker_service(slot: int) -> str:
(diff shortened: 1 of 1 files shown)


msilvafe:
I definitely see @BrianJKoopman's frustration with the current setup so thanks for addressing this but can you describe what testing you've done before I approve?

jlashner:
Ok, I moved the operations for starting / stopping the pysmurf-controllers to a separate function so that they could be called and tested outside of a full hammer.

I also added the subcommand `jackhammer controller <up/down> <slots>`  to make it easier to manage pysmurf-controllers without looking up their instance-ids.

This has been tested on satp1.

msilvafe (approved):
Great, thanks for testing!

PR #449, opened 2024-12-10
Support usage of SMURF_CONFIG_DIR instead of OCS_CONFIG_DIR

jlashner:
This PR adds support for an environment variable $SMURF_CONFIG_DIR which will replace $OCS_CONFIG_DIR if it is specified.

The OCS_CONFIG_DIR technically needs to be the directory where the OCS default.yaml file exists, which for most lab-setups is the same as the directory where the sys_config.yml and smurf-config files live.  However for deployed servers these are different. We've been getting around by always manually specifying the ocs site-config file on deployment servers, but this will no longer be possible when the smurf-server environment is  managed more by site-computing.

Moving forward, we will want to point to the smurf-configuration directory using a separate environment variable, $SMURF_CONFIG_DIR, which defaults to OCS_CONFIG_DIR if it is not set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,4 @@ venv.bak/
 
 # vscode
 .vscode
+.devcontainer/
--- a/docs/configs.rst
+++ b/docs/configs.rst
@@ -5,10 +5,17 @@ SMuRF is a complex system with many configuration parameters which vary
 depending on the cryostat, the experiment being conducted, and the network
 configuration of the smurf-server and data-acquisition nodes.
 
-These config parameters will be stored in various files in the
-``OCS_CONFIG_DIR``, where institutions will maintain and version-control the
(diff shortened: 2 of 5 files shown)


BrianJKoopman (commented):
This looks good to me, but doing a `git grep 'OCS_CONFIG_DIR'` I noticed some other spots that probably could use updating to use `SMURF_CONFIG_DIR`:
* `docs/configs.rst`
* `docs/scripting_guide.rst`
* `tests/input/sys_config.yml`

jlashner:
Thanks Brian, I clarified the docs with regards to the configuration directory and changed wording to be more general.

BrianJKoopman (approved):

jlashner:
This has been tested with the new deployment configuration on smurf-srv20-satp1.

PR #451, opened 2025-01-22
Hostmanager + jackhammer compatibility

jlashner:
In the near future, we want to get hostmanager operating on smurf-servers to automatically start ocs agents.

Once this is the case, we will not want to touch ocs dockers using jackhammer... it should all be managed by the hostmanager.

With this PR, if  `sys_config["use_hostmanager"]` is set to True, it will exclude ocs dockers from any hammer operations. 

When setting this up, we will need to move ocs-related docker services to an independent docker-compose file to keep host-manager from interacting with smurf-streamer dockers, as is done here: https://github.com/simonsobs/ocs-deployment-configs/pull/332.

### DIFF
--- a/hammers/jackhammer
+++ b/hammers/jackhammer
@@ -76,7 +76,7 @@ def get_docker_services():
     Returns a list of docker services that are available from the docker-compose
     file in the ocs config directory.
     """
-    cmd = 'docker-compose config --services'
+    cmd = 'docker compose config --services'
     res = subprocess.run(shlex.split(cmd), cwd=cwd, stdout=subprocess.PIPE)
     return res.stdout.decode().split()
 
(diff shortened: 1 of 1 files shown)


jlashner:
I've tested this with satp3 and its good to merge!

One thing is that there was an issue (https://github.com/simonsobs/ocs/issues/425) that required us to change all `docker-compose` commands to `docker compose`.

The old `docker-compose` v1 has been deprecated for several years so I think this is ok. I believe should only effect servers running ubuntu 18.04, so nothing in operation, and any lab-setup running 18.04 should probably upgrade anyway.

msilvafe (approved):
We need to send out some sort of slack blast to smurf users to let them know that docker-compose is no longer used but I think its time. This change happened a while ago now

PR #454, opened 2025-02-12
Adds some error handling needed for atten optimization on the LAT

jlashner:
This error handling was needed when optimizing attenuation on the LAT. Most likely because re-tuning between steps occasionally drops channels.


tpsatt:
Looks good to me. Thanks!

tpsatt (approved):

PR #455, opened 2025-02-24
clear_data: Add --force arg to skip confirmation

jlashner:
Adds option to run data deletion script without user input prompt, useful for setting up automated cronjobs.

Tested this on smurf-srv21-satp2.


msilvafe (approved):
Thanks for getting this running.

PR #457, opened 2025-03-04
feat(stream.streamg3_on): Set filter parameters.

tristpinsm:
I've yet to test this on a SMuRF system (working on getting that going), and that will involve confirming that I've correctly set the default filter parameters, but I figured it would be useful in the meantime to get feedback on how I've implemented this.

### DIFF
--- a/sodetlib/stream.py
+++ b/sodetlib/stream.py
@@ -130,8 +130,8 @@ def take_g3_data(S, dur, **stream_kw):
 def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
                  channel_mask=None, filter_wait_time=2, make_datfile=False,
                  downsample_factor=None, downsample_mode=None,
-                 filter_disable=False, stream_type=None, subtype='stream',
-                 enable_compression=None):
+                 filter_disable=False, filter_order=None, filter_cutoff=None,
+                 stream_type=None, subtype='stream', enable_compression=None):
     """
     Starts the G3 data-stream. Returns the session-id corresponding with the
     data stream.
@@ -162,6 +162,11 @@ def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
         this will be pulled from the device cfg.
     filter_disable : bool
         If true, will disable the downsample filter before streaming.
+    filter_order : int
+        Order of the downsample filter. Read from device config by default.
+    filter_cutoff : float
+        The cutoff frequency in Hz for the downsample filter.
+        Read from device config by default.
     stream_type : optional(string)
         Type of stream. This should either be "obs" or "oper". If None,
         it will be determined based off of the subtype. ('stream', 'cmb', and
@@ -202,9 +207,14 @@ def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
         downsample_mode = cfg.dev.exp['downsample_mode']
     if downsample_factor is None:
         downsample_factor = cfg.dev.exp['downsample_factor']
+    if filter_order is None:
+        filter_order = cfg.dev.exp["downsample_filter"]["order"]
+    if filter_cutoff is None:
+        filter_cutoff = cfg.dev.exp["downsample_filter"]["cutoff_freq"]
 
     S.set_downsample_mode(downsample_mode)
     S.set_downsample_factor(downsample_factor)
+    S.set_downsample_filter(filter_order, filter_cutoff)
     S.set_filter_disable(int(filter_disable))
 
     S.stream_data_on(make_freq_mask=make_freq_mask, channel_mask=channel_mask,
--- a/sodetlib/det_config.py
+++ b/sodetlib/det_config.py
@@ -42,6 +42,10 @@ exp_defaults = {
     'downsample_mode': 'internal',
     'enable_compression': True,
 
+    # Downsample filter parameters
+    # cutoff freq in Hz reverse engineered to produce pysmurf default filter
+    "downsample_filter": {"order": 4, "cutoff_freq": 63.0},
+
     # Dict of device-specific default params to use for take_iv
     'iv_defaults': {},
 


sodetlib/stream.py
```diff
@@ -202,9 +205,12 @@ def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
         downsample_mode = cfg.dev.exp['downsample_mode']
     if downsample_factor is None:
         downsample_factor = cfg.dev.exp['downsample_factor']
+    if filter_params is None:
+        filter_params = cfg.dev.exp["downsample_filter"]
 
     S.set_downsample_mode(downsample_mode)
     S.set_downsample_factor(downsample_factor)
+    S.set_downsample_filter(filter_params["order"], filter_params["cutoff_freq"])
```
jlashner, line 213: So ideally, we don't want this PR to change default filter behavior. The default filter parameters are defined here: 
https://github.com/simonsobs/pysmurf/blob/dc73e180dab566f6ed2812bb1ba775c2c1a32232/python/pysmurf/core/devices/_SmurfProcessor.py#L155-L181

I think the importance of this is kind of determined by how this is handled by sotodlib, so we should talk to Matthew Hasselfield about it. If we can find the filter_order and filter_cutoff used to produce approximately the same filter, and the pipeline doesn't care much about small changes in these filter parameters, then I think setting this using order/cutoff is fine.

However if we want the filter parameters to match exactly what we've been using, then we should put the default parameters into the device cfg as something like `filter_a` and `filter_b`, and set directly if no cutoff freq is specified.

sodetlib/stream.py
```diff
@@ -130,8 +130,8 @@ def take_g3_data(S, dur, **stream_kw):
 def stream_g3_on(S, make_freq_mask=False, emulator=False, tag=None,
                  channel_mask=None, filter_wait_time=2, make_datfile=False,
                  downsample_factor=None, downsample_mode=None,
-                 filter_disable=False, stream_type=None, subtype='stream',
-                 enable_compression=None):
+                 filter_disable=False, filter_params=None, stream_type=None,
```
jlashner, line 133: I think we should specify `filter_order` and `filter_cutoff` separately instead of passing in a dict. I think usually we'll want to just change the filter_cutoff while using the default filter_order, and how its currently written means you always need to pass both together.

jlashner (changes requested):
Hi Tristan, thanks for this! It looks good for the most part, see comments below. Mainly I think we need to talk with Matthew / Max about whether its ok to start specifying filter behavior using `filter_cutoff` if that means the default filter parameters don't match exactly what's currently used.

We'll also want to investigate what filter cutoff was used to generate the default parameters that we currently use.

tristpinsm:
I've confirmed that the filter parameters in the device config produce the default coefficients on a live SMuRF system now.

The coefficients queried before running `S.set_downsample_filter(filter_params["order"], filter_params["cutoff_freq"])`
```
[ 1.         -3.74145562  5.25726624 -3.28776591  0.77203984  0.
  0.          0.          0.          0.          0.          0.
  0.          0.          0.          0.        ]
[5.28396689e-06 2.11358676e-05 3.17038014e-05 2.11358676e-05
 5.28396689e-06 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00]
```
and after
```
[ 1.         -3.74145562  5.25726624 -3.28776591  0.77203984]
[5.28396689e-06 2.11358676e-05 3.17038014e-05 2.11358676e-05
 5.28396689e-06]
```
I also checked that they are different if I change the cutoff frequency :)

tristpinsm:
@swh76 may have an opinion on this change?

jlashner:
Great. We should also check with shawn / make sure we test this, but I _believe_ its ok that we don't pad with zeros when we update the filter params. I think that is just done so the initial buffer is set to a large enough length.

tristpinsm:
@swh76 @msilvafe Should we try to get this merged before we release a new software stack with the RSSI fix?

msilvafe:
> @swh76 @msilvafe Should we try to get this merged before we release a new software stack with the RSSI fix?

I'm happy with it, remind me how you've tested it so far?

tristpinsm:
> I'm happy with it, remind me how you've tested it so far?

A few comments upstream I checked that the parameters were exactly the same as before using the default values stored in the config. I haven't looked at any data, do you think that is necessary?

msilvafe:
> > I'm happy with it, remind me how you've tested it so far?
> 
> A few comments upstream I checked that the parameters were exactly the same as before using the default values stored in the config. I haven't looked at any data, do you think that is necessary?

I think what you checked above seems good. But I have Jack's same uncertainty about the zero padding. Perhaps to stay consistent we zero pad the arrays?

tristpinsm:
> > > I'm happy with it, remind me how you've tested it so far?
> > 
> > 
> > A few comments upstream I checked that the parameters were exactly the same as before using the default values stored in the config. I haven't looked at any data, do you think that is necessary?
> 
> I think what you checked above seems good. But I have Jack's same uncertainty about the zero padding. Perhaps to stay consistent we zero pad the arrays?

I think this is just a quirk of EPICS (see [here](https://github.com/slaclab/pysmurf/blob/8320dce3609c9e1f76c45ff4a34b0d1e26b33643/python/pysmurf/core/devices/_SmurfProcessor.py#L181)). The actual filter doesn't care about the length ([here](https://github.com/slaclab/pysmurf/blob/8320dce3609c9e1f76c45ff4a34b0d1e26b33643/src/smurf/core/processors/SmurfProcessor.cpp#L294)).

The fact I was able to set it and it didn't complain suggests that EPICS is happy without zero-padding.

Would still be nice to get @swh76 opinion.

tristpinsm:
Should we go ahead and merge? In any case, one reviewer will need to approve :)

msilvafe (approved):
Sorry to hold ya up!

PR #461, opened 2025-04-23
Remove logging of every bias step.

tristpinsm:
During my SMuRF tsar shift I noticed that the logs in the pysmurf controllers were often hard to read due to their volume and the mix of `sodetlib` / `pysmurf` / other origins that were sometimes conflicting. I think there might be some improvement possible on the `pysmurf` side, so I'll create an issue there.

This was one particularly egregious example.

### DIFF
--- a/sodetlib/operations/iv.py
+++ b/sodetlib/operations/iv.py
@@ -646,7 +646,7 @@ def take_iv(S, cfg, **kwargs):
         S.log(f"Starting TES Bias Ramp on bg {bgs}")
         for i, bias in enumerate(ivcfg.biases):
             sdl.stop_point(S)
-            S.log(f"Setting bias to {bias:4.3f}")
+            S.log(f"Setting bias to {bias:4.3f}", S.LOG_INFO)
             _bias_arr[bgs] = bias
             S.set_tes_bias_bipolar_array(_bias_arr)
             start_times[bgs, i] = time.time()


msilvafe (commented):
I looked into the smurf logger class and couldn't tell exactly how the log verbosity works but do you think it's possible to log this at a higher log level like "debug" in a standard python logger instead of deleting it directly?

msilvafe:
> Yeah, it's not super easy to find, but the log verbosity is defined [here](https://github.com/slaclab/pysmurf/blob/8320dce3609c9e1f76c45ff4a34b0d1e26b33643/python/pysmurf/client/base/base_class.py#L44). It looks like by default it logs at level 0 (least verbose).
> 
> So I could set it to `LOG_INFO` (1), which (should) suppress these in the `pysmurf_controller` logs. But to confirm I'd have to test it out I think.

Ok, is this something your could test on the slac system?

tristpinsm:
what would be easy to do is put in some log statements and play with the verbosity to confirm it behaves as expected.

tristpinsm:
> what would be easy to do is put in some log statements and play with the verbosity to confirm it behaves as expected.

OK I've done that and don't see messages logged at `S.LOG_INFO` level.

msilvafe (approved):
Ok, great. Not an issue for this PR but how do we change the log levels for the pysmurf logger if someone wanted to get the bias step messages back?

tristpinsm:
> Ok, great. Not an issue for this PR but how do we change the log levels for the pysmurf logger if someone wanted to get the bias step messages back?

It's not very obvious, but you can set it with `verbose` argument to the smurf client object instantiation, or set it later using the `set_verbose` method. e.g. [here](https://github.com/slaclab/pysmurf/blob/ac16235027ccb10720b471842f1bbfae9a353c32/python/pysmurf/client/base/base_class.py#L71)

PR #460, opened 2025-04-10
remove docker container.

tristpinsm:
I'm proposing we remove the `sodetlib` docker container. As far as I can tell it is only used to build the `ocs-pysmurf-agent` ([here](https://github.com/simonsobs/socs/blob/main/docker/pysmurf_controller/Dockerfile)), and we could simply install `sodetlib` and specify a release tag there. The only other reference I have found to this image are in `docker-compose.yaml` files committed to the `ocs-site-configs` repo for some institutions. In the SAT/LAT configs it has been replaced by `ocs-pysmurf-agent`. We should then also remove the `SODETLIB_TAG` that remains in those configs but is unused.

This would simplify the release process by removing one layer of docker inheritance, but I appreciate that I may have overlooked other uses of this image (although I did not find any in the `simonsobs` organisation), so please let me know if you think this may be the case.

An accompanying PR will be necessary in `socs` to install `sodetlib` in the corresponding `Dockerfile`.

PR #470, opened 2025-08-28
Remove SODETLIB_TAG from config file docs

BrianJKoopman:
We stopped building the sodetlib image and started installing it directly into the pysmurf controller agent image. See [1] for the removal details.

[1] - https://github.com/simonsobs/sodetlib/pull/460


msilvafe:
Ok with me, @tristpinsm have any issues?

PR #474, opened 2025-10-15
Perform bg_map at end of relock

tpsatt:
Responds to problem where `bg_map` wasn't run after every (potentially polarity-changing) relock.


dpdutcher (approved):
Tested on a UFM in the Princeton testbed, and it works as expected.

PR #485, opened 2025-12-10
Updates to support rogue 6

tristpinsm:
Tested on SLAC system

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ scipy
 matplotlib
 lmfit
 pandas
+zmq
 # Can install sotodlib from PyPI once Python version in pysmurf-controller is >=3.10
 # i.e. when [1] is merged and propagated to [2].
 # [1] - https://github.com/simonsobs/smurf_dockers/pull/6
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,11 +6,7 @@ matplotlib
 lmfit
 pandas
 zmq
-# Can install sotodlib from PyPI once Python version in pysmurf-controller is >=3.10
-# i.e. when [1] is merged and propagated to [2].
-# [1] - https://github.com/simonsobs/smurf_dockers/pull/6
-# [2] - https://github.com/simonsobs/socs/blob/main/docker/pysmurf_controller/Dockerfile
-sotodlib @ git+https://github.com/simonsobs/sotodlib.git@5d613d5915b1716c401abecb5446088bce5fc1a4
+sotodlib
 pysmurf-slac
 
 # scripts
--- a/sodetlib/__init__.py
+++ b/sodetlib/__init__.py
@@ -1,7 +1,6 @@
 import os
 from functools import wraps
 try:
-    import epics
     from pysmurf.client.util.pub import set_action
 except Exception as e:
     # Just return base function regularly if can't import set_action
--- a/sodetlib/hammers/jackhammer.py
+++ b/sodetlib/hammers/jackhammer.py
@@ -255,7 +255,7 @@ def dump_docker_logs(slots, dump_rogue_tree=False):
     if dump_rogue_tree:
         dump_script = '/sodetlib/scripts/dump_rogue_state.py'
         for slot in slots:
-            if check_server_connection(9000 + 2 * slot, retry=False):
+            if check_server_connection(9000 + 3 * slot, retry=False):
                 out_file = os.path.join(dump_dir, f'rogue_state_s{slot}.yml')
                 cprint(f"Dumping s{slot} state to {out_file}", style=TermColors.HEADER)
                 util_run('python3', args=[dump_script, str(slot), out_file],
@@ -486,7 +486,7 @@ def hammer_func(args):
     # Waits for streamer-dockers to start
     print("Waiting for server dockers to connect. This might take a few minutes...")
     for slot in slots:
-        check_server_connection(9000 + 2 * slot, retry=True)
+        check_server_connection(9000 + 3 * slot, retry=True)
 
     if reboot and not args.skip_setup:
         cprint("Configuring pysmurf", style=TermColors.HEADER)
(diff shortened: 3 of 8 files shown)


sodetlib/hammers/jackhammer.py
```diff
@@ -265,13 +282,13 @@ def dump_docker_logs(slots, dump_rogue_tree=False):
     if dump_rogue_tree:
         dump_script = '/sodetlib/scripts/dump_rogue_state.py'
         for slot in slots:
-            if check_epics_connection(f'smurf_server_s{slot}', retry=False):
+            if check_server_connection(9000 + 3 * slot, retry=False):
```
BrianJKoopman: Just curious, is there somewhere this `9000 + 3 * slot` for the server port is documented? Is this a new Rogue 6 ZMQ thing that's documented in pysmurf somewhere maybe?

requirements.txt
```diff
@@ -5,12 +5,9 @@ scipy
...
 lmfit
 pandas
-# Can install sotodlib from PyPI once Python version in pysmurf-controller is >=3.10
-# i.e. when [1] is merged and propagated to [2].
-# [1] - https://github.com/simonsobs/smurf_dockers/pull/6
-# [2] - https://github.com/simonsobs/socs/blob/main/docker/pysmurf_controller/Dockerfile
-sotodlib @ git+https://github.com/simonsobs/sotodlib.git@5d613d5915b1716c401abecb5446088bce5fc1a4
-pysmurf-slac==10.3.4
+zmq
+sotodlib
```
BrianJKoopman: Great, I'd also add both of these to the dependencies list in `pyproject.toml`. (`sotodlib` wasn't in there before because of the need to install from VCS, but that's resolved now that we're on 3.12 in the container.)

sodetlib/__init__.py
BrianJKoopman: Commenting here for lack of a better place -- does this PR totally eliminate the `epics` dependency?

If so, there are a couple of lingering spots it's mentioned/used:
```
$ git grep -e 'epics'
docs/det_config.rst:DetConfig object will automaticlly determine the epics root and set the
scripts/resonator_fitting.py:   parser.add_argument('--epics-root', default='smurf_server_s2')
scripts/resonator_fitting.py:                   epics_root = args.epics_root,
scripts/restart_rssi.py:import epics
scripts/restart_rssi.py:ipbyeng=[epics.PV(name).get(as_string=True) for name in [client_remote_ip0.format(iudpe) for iudpe in range(nudpe)]]
scripts/restart_rssi.py:    validcntpv=epics.PV(rssi_validcnt.format(slots[slot]['engine']))
scripts/restart_rssi.py:    restartconnpv=epics.PV(rssi_restartconn.format(slots[slot]['engine']))
scripts/set_stream.py:    parser.add_argument('--epics-root', type=str, default=None)
scripts/set_stream.py:        dump_configs=args.dump_configs, epics_root=args.epics_root,
sodetlib/util.py:    existing pysmurf function so we can set both PV's in a single epics call,
tests/input/pysmurf.cfg:"epics_root" : "test_epics",
tests/test_jackhammer.py:                    f"connection to {epics_server}"
```

I think at least `tests/test_jackhammer.py`, the description in `sodetlib/util.py`, and the docs in `docs/det_config.rst` should probably be addressed. The `scripts/` directory I'm less sure what to do about. If dropping epics in those scripts is straight forward, let's do that too.

`tests/input/pysmurf.cfg` and generally the `tests/test_det_config.py` file looks like it needs attention -- it's not passing for me on a fresh install from `master` or from this branch. That doesn't have to happen in this PR though.

BrianJKoopman (commented):
This looks good to me, with the caveat that I'm not really familiar enough with the epics/zmq details and inner pysmurf stuff to really comment on the changes in the operations. My comments focus on the packaging, and I'm wondering about the `epics` dependency.

sodetlib/hammers/jackhammer.py
tristpinsm: it's a convention that carries over from the old code, which actually did have a ZMQ server running, it just wasn't used for the rogue tree.

As in typical `pysmurf` fashion is it documented in docstrings / comments in the code.

There's a longer term project to refactor all of this, but for now I can make sure this gets documented in the compiled `pysmurf` docs.

requirements.txt
tristpinsm: I was wondering about this earlier: why does this list need to be in two places? It bit me in this branch when I merged in the upstream `main` branch and only removed the pinned `pysmurf` version in `requirements.txt` but not `pyproject.toml`. (I'm sure there's a good reason, just looking to understand why)

sodetlib/__init__.py
tristpinsm: that's a good point -- the `scripts` directory slipped by me. Shouldn't be too hard to adapt those. Are those tests configured to run? I haven't seen them in CI.

Thanks for taking a look!

sodetlib/hammers/jackhammer.py
tristpinsm: added to the docs in slaclab/pysmurf#1047

requirements.txt
BrianJKoopman: Yeah, so the list in `pyproject.toml` is what defines the dependencies that are automatically installed when you `pip install .` the `sodetlib` package (even if you install from PyPI.) The `requirements.txt` file requires the explicit `pip install -r requirements.txt`.

The `pyproject.toml` list doesn't support installing from VCS, so you can't list something like:
```
sotodlib @ git+https://github.com/simonsobs/sotodlib.git@5d613d5915b1716c401abecb5446088bce5fc1a4
```

The requirements file does support this -- so we recommended an install using the requirements file in order to install all the dependencies in one command.

Now that `sotodlib` is properly packaged and published to PyPI, and we've listed all the dependencies in the `pyproject.toml` list, we can probably drop the `requirements.txt` file altogether. (There may still be uses for it, but I don't think we really have one any more.)

sodetlib/__init__.py
BrianJKoopman: Great, I think there are some cases where the `scripts/` are still used -- though I'd be happy to be wrong about that. (They're sort of awkward here, since they're not *really* a part of the package that gets published to PyPI, but last I looked I think `jackhammer` interacted with that directory in some way.)

Also, no, there's no CI to run those tests (which is probably why the breakage went without anyone noticing). If you want to set that up (either here or in a new PR), I'd be happy to review that too. There aren't many tests in this repo, so I think motivation was low.

PR #501, opened 2026-02-12
Install pysmurf from PyPI

BrianJKoopman:
Now that pysmurf is [published to PyPI](https://pypi.org/project/pysmurf-slac/), we can install the latest release from there instead of needing to install from VCS.

This doesn't currently change anything downstream in the pysmurf-controller Docker image. The `pysmurf-slac` package gets installed (version 9.0.1), but the version 8.1.0 that is cloned and [inserted into the `PYTHONPATH`](https://github.com/simonsobs/smurf_dockers/blob/3bb520cca801d46981af063309ab1d5f94503b83/pysmurf/Dockerfile#L37) takes precedence and is what ends up being used. This'll be true until https://github.com/simonsobs/smurf_dockers/issues/7 is resolved.

That said, happy to wait on this until the Rogue 6 update happens. Whatever is easiest for you @tristpinsm.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "matplotlib",
     "numpy",
     "pandas",
+    "pysmurf-slac",
     "PyYAML",
     "scipy",
     "tqdm",
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,9 +10,7 @@ pandas
 # [1] - https://github.com/simonsobs/smurf_dockers/pull/6
 # [2] - https://github.com/simonsobs/socs/blob/main/docker/pysmurf_controller/Dockerfile
 sotodlib @ git+https://github.com/simonsobs/sotodlib.git@5d613d5915b1716c401abecb5446088bce5fc1a4
-# pysmurf can be installed from PyPI once [3] is merged and a new release is made
-# [3] - https://github.com/slaclab/pysmurf/pull/824
-pysmurf @ git+https://github.com/slaclab/pysmurf.git@main
+pysmurf-slac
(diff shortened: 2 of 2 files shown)


BrianJKoopman:
I did some more testing and now understand the interaction between this and how pysmurf is installed currently in the `simonsobs/so_smurf_base` base image. This PR, while it will end up pulling in pysmurf 9.0.1, version 8.1.0 will still be inserted at the front of the `PYTHONPATH`, and will get used instead until https://github.com/simonsobs/smurf_dockers/issues/7 is resolved. Marked ready for review, though as I edited in the main PR text, no rush, happy to wait until the Rogue 6 update.

tristpinsm:
Yeah, the way things are built currently relies on modifying the `PYTHONPATH` explicitly... I think it would be better to move away from that, but it would require some testing because I'm not sure whether anything downstream will be affected by such a change.

We will still need to check out the pysmurf source in those dockers unless start releasing compiled libraries via pip.

tristpinsm (approved):
Despite this not making a difference in the SO docker installs, it still makes sense to update the requirements I think :)

BrianJKoopman:
> Yeah, the way things are built currently relies on modifying the `PYTHONPATH` explicitly... I think it would be better to move away from that, but it would require some testing because I'm not sure whether anything downstream will be affected by such a change.
> 
> We will still need to check out the pysmurf source in those dockers unless start releasing compiled libraries via pip.

Yeah, I don't quite understand the interaction between the server and client code when compiling the server code. But also I don't really want to tackle packaging the compiled libraries either...

Issue #473, opened 2025-10-01
ASU 4K LNAs need different biasing approach


dpdutcher:
The 4K LNAs produced by ASU we're using for the ASO LATR OTs require a different biasing scheme than the LNF LNAs. Specifically, ASU LNAs require gate voltage = 0 and they also have different safe input ranges compared to the LNF.

* LNF
  * (min,max) values
    * Vd: (-0.5, 2) V
    * Vg: (-20, 20) V
  * optimal 4K bias
    * Vd: 0.8 V
    * Vg: -1.18 V
    * Id: 11 mA 
* ASU
  *  (min, max) values
    * Vd: (0, 1) V
    * Vg: (-1, 0) V
  * optimal 4K bias
    * Vd: 0.8 V
    * Vg: 0 V
    * Id: 3--7 mA  

This is a must-fix, and requires changes to the config file to specify which type of 4K amp is in use, and then changes to setup_amps to bias the different amplifiers differently.

dpdutcher:
Talking with Justin at ASU, it sounds like we can just set Vd=0.8, Vg=0 and be done with it. No adjusting Vg.

dpdutcher:
We were successfully biasing ASU 4K amps in the LAT before it started warm-up, so I'm declaring this solved.

PR #502, opened 2026-02-19
Scan over drain voltage for ASU HEMTs

dpdutcher:
Addresses #473 

This checks for ASU 4K hemt amplifiers by evaluating the min,max values of Vg in the device config file. If both of those are set to 0, then setup_amps scans over drain voltage instead of gate voltage.  This is not foolproof by any means, but any improper config files need to be addresed anyway.

This was tested successfully on a cold system on the LAT (c5s2) connected to ASU 4K and 50K amplifiers.


msilvafe (approved):
Approved assuming you've tested.

PR #508, opened 2026-06-11
Wrap setup_amps in try during relock for two attempts

tpsatt:
Semi-frequently after a hammer `setup_amps` fails on `uxm_relock`, usually due to a key error for the name of one of the amplifiers. It's pretty much always fixed by running `uxm_relock` again, but ROCs don't know to do this and smurf-tsars can be slow to respond.

This is a very simple solution of wrapping `setup_amps` in a `try`. If it fails, it tries again. If the second attempt fails, then it errors out.

### DIFF
--- a/sodetlib/operations/uxm_relock.py
+++ b/sodetlib/operations/uxm_relock.py
@@ -262,7 +262,7 @@ def plot_channel_resonance(S, cfg, band, chan):
 def uxm_relock(
     S, cfg, bands=None, show_plots=False,
     setup_notches=False, new_master_assignment=False, reset_rate_khz=None,
-    nphi0=None, skip_setup_amps=False):
+    nphi0=None, skip_setup_amps=False, max_setup_amps_attempts=3):
     """
     Relocks resonators by running the following steps:
 
@@ -294,6 +294,8 @@ def uxm_relock(
         If True will show plots
     skip_setup_amps : bool
         If True will skip amplifier setup.
+    max_setup_amps_attempts : int
+        Maximum number of attempts at setting up amplifiers before returning error
 
     Returns
     --------
@@ -335,15 +337,13 @@ def uxm_relock(
     if not skip_setup_amps:
         summary['timestamps'].append(('setup_amps', time.time()))
         sdl.set_session_data(S, 'timestamps', summary['timestamps'])
-
-        # May take two tries
-        try:
-            success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
-        except:
+        for i in range(max_setup_amps_attempts):
             try:
                 success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+                break
             except:
-                return False
+                success = False
+                S.log('Failed uxm_setup.setup_amps on %i/%i attempt' % (i, max_setup_amps_attempts))
 
         if not success:
             return False, summary
--- a/sodetlib/operations/uxm_relock.py
+++ b/sodetlib/operations/uxm_relock.py
@@ -335,7 +335,15 @@ def uxm_relock(
     if not skip_setup_amps:
         summary['timestamps'].append(('setup_amps', time.time()))
         sdl.set_session_data(S, 'timestamps', summary['timestamps'])
-        success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+
+        # May take two tries
+        try:
+            success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+        except:
+            try:
+                success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+            except:
+                return False
 
         if not success:
             return False, summary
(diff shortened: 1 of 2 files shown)


sodetlib/operations/uxm_relock.py
```diff
@@ -335,7 +335,15 @@ def uxm_relock(
...
-        success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+
+        # May take two tries
+        try:
+            success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+        except:
+            try:
+                success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+            except:
+                return False
```
tristpinsm:
```suggestion
                return False, summary
```

sodetlib/operations/uxm_relock.py
```diff
@@ -335,7 +335,15 @@ def uxm_relock(
     if not skip_setup_amps:
         summary['timestamps'].append(('setup_amps', time.time()))
         sdl.set_session_data(S, 'timestamps', summary['timestamps'])
-        success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+
+        # May take two tries
+        try:
+            success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+        except:
```
tristpinsm:
```suggestion
        except:
            time.sleep(0.1)
```
Just a bit of insurance in case super rapid back-to-back calls cause issues.

tristpinsm (changes requested):
Not loving the nested `try/except` here but ok :)

tristpinsm (approved):

tristpinsm (changes requested):
Actually, sorry I approved too soon. One more request: Could we log the exception when it is caught? Otherwise it gets really hard to figure out what is going on.

sodetlib/operations/uxm_relock.py
```diff
@@ -335,7 +337,14 @@ def uxm_relock(
     if not skip_setup_amps:
         summary['timestamps'].append(('setup_amps', time.time()))
         sdl.set_session_data(S, 'timestamps', summary['timestamps'])
-        success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
+        for i in range(max_setup_amps_attempts):
```
tristpinsm: It looks like [the pysmurf_controller](https://github.com/simonsobs/socs/blob/88d7bf40561f4f16aa5c775d2f4f9496d74fbdfd/socs/agents/pysmurf_controller/smurf_subprocess_util.py#L340) ignores the success return value. For it to register a failure the exception has to be raised.

```suggestion
        i = 0
        success = False
        while not success:
            i += 1
            try:
                success, summary['amps'] = uxm_setup.setup_amps(S, cfg)
            except Exception as e:
                S.log('Failed uxm_setup.setup_amps on %i/%i attempt' % (i, max_setup_amps_attempts))
                if i < max_setup_amps_attempts:
                    # sleep and continue trying
                    time.sleep(0.1)
                else:
                    raise e
```

dpdutcher (changes requested):
Can this logic be moved to be within `uxm_setup.setup_amps` instead? 
Here: https://github.com/simonsobs/sodetlib/blob/279def4e2d9c2e7dfd4e22fe805b8e8362cc0477/sodetlib/operations/uxm_setup.py#L217-L222

That's the cause of the failure and where it should be caught and dealt with, not in this function. uxm_setup.uxm_setup also has a call to `setup_amps`, and repeating this check there would duplicate code.

sodetlib/operations/uxm_setup.py
```diff
@@ -214,7 +217,20 @@ def setup_amps(S, cfg, update_cfg=True, enable_300K_LNA=True):
...
 
-    amp_biases = S.get_amplifier_biases()
+    attempts = 0
+    read_biases_success = False
+    while not read_biases_success:
+        attempts += 1
+        try:
+            amp_biases = S.get_amplifier_biases()
+            read_biases_success = True
+        except Exception as e:
```
dpdutcher, line 232: The failure mode won't necessarily be an exception thrown by get_amplifier_biases, I think the more common error is that it returns an empty dictionary.

tristpinsm (approved):
looks good to me

dpdutcher (approved):

sodetlib/operations/uxm_setup.py
```diff
@@ -214,7 +217,25 @@ def setup_amps(S, cfg, update_cfg=True, enable_300K_LNA=True):
...
+    attempts = 0
+    read_biases_success = False
+    while not read_biases_success:
+        attempts += 1
+        try:
+            amp_biases = S.get_amplifier_biases()
+
+            # Attempt to read Vds
+            for amp in amp_list:
+                _ = amp_biases[f"{amp}_drain_volt"]
```
dpdutcher: This may seem like a niche complaint, but the _drain_volt keys are only returned for cryocards C04+.
Now, every cryocard in use in SO that I'm aware of is C05, and certainly that's required for use at the site, though its not impossible there are some C02s still at NA institutions. Checking _drain_current or _gate_volt would be more cryocard version-agnostic, or one could simply check that the dictionary isn't empty.

PR #514, opened 2026-07-22
Expose jackhammer function to programmatic use

tristpinsm:
Needed for the `socs` jackhammer agent.

Allows the script to complete even if individual slots fail.


msilvafe (approved):
This looks good to me, and you added tests! Let me know how the LAT testing goes.

PR #516, opened 2026-08-26
Fix bugs in jackhammer

tristpinsm:
1. Parts of the script are run in docker containers. This allocates a TTY by default which messes with standard output afterwards. Disable this except for interactive sessions.
2. Remove mutable defaults trap. fixes #515


msilvafe (approved):
Looks good. I love when the python documentation calls me a neophyte.

BrianJKoopman (commented):
Looks good! Just one small suggestion, which you can change if you'd like, otherwise merge away.

BrianJKoopman (approved):

